PR #55, opened 2021-03-31
Add CompoundWrapAdapter

cgewecke:
(Continuation of [set-protocol-v2-contracts PR 19][1])

* Adds Compound wrapper module

[1]: https://github.com/SetProtocol/set-protocol-v2-contracts/pull/19

### DIFF
--- a/test/integration/compoundWrapModule.spec.ts
+++ b/test/integration/compoundWrapModule.spec.ts
@@ -1,0 +1,200 @@
+import "module-alias/register";
+import { BigNumber } from "@ethersproject/bignumber";
+
+import { Address } from "@utils/types";
+import { Account } from "@utils/test/types";
+import { ADDRESS_ZERO } from "@utils/constants";
+import { CompoundWrapAdapter, SetToken, WrapModule } from "@utils/contracts";
+import { CERc20 } from "@utils/contracts/compound";
+import DeployHelper from "@utils/deploys";
+import {
+  ether,
+  preciseMul,
+} from "@utils/index";
+import {
+  getAccounts,
+  getWaffleExpect,
+  getSystemFixture,
+  getCompoundFixture,
+  addSnapshotBeforeRestoreAfterEach,
+} from "@utils/test/index";
+import { CompoundFixture, SystemFixture } from "@utils/fixtures";
+
+const expect = getWaffleExpect();
+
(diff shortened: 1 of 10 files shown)


test/integration/compoundWrapModule.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+      it("should reduce the underlying quantity and mint the wrapped asset to the SetToken", async () => {
+        const previousUnderlyingBalance = await setup.dai.balanceOf(setToken.address);
+
+        const previousWrappedBalance = await cDai.balanceOf(setToken.address);
+
+        await subject();
+
+        const underlyingBalance = await setup.dai.balanceOf(setToken.address);
+        const wrappedBalance = await cDai.balanceOf(setToken.address);
+        const expectedUnderlyingBalance = previousUnderlyingBalance.sub(setTokensIssued);
```
cgewecke, line 137: Do these tests pass? Seems like there would be a mismatch of decimals since setTokens are `10**18` and cTokens are `10**8`

_Originally posted by @bweick in https://github.com/SetProtocol/set-protocol-v2-contracts-legacy/pull/19#discussion_r599101694_

cgewecke: It works because it is subtracting DAI and Set Token, both have 10**18

_Originally posted by @MarioCerdan in https://github.com/SetProtocol/set-protocol-v2-contracts-legacy/pull/19#discussion_r599209303_

cgewecke:
So we have integration tests (i.e. module using the adapter) but think we also want direct unit tests that check to make sure the correct calldata is being returned. Those tests would go in the `test/protocol/integration` folder (whereas the integration tests you did were correctly placed in the `test/integration` folder).

_Originally posted by @bweick in https://github.com/SetProtocol/set-protocol-v2-contracts-legacy/issues/19#issuecomment-804423748_

test/integration/compoundWrapModule.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+
+        expect(underlyingBalance).to.eq(expectedUnderlyingBalance);
+
+        const underlyingTokenDecimals = await setup.dai.decimals();
```
bweick, line 141: These two variables seem to be unused

bweick (approved):

PR #59, opened 2021-04-01
Add `GeneralIndexModule` contract

Alpha-Guy:
- Add `GeneralIndexModule` contract
- Add `BalancerV1ExchangeAdapter` contract
- Refactor `UniswapV2ExchangeAdapterV2` contract


contracts/protocol/integration/UniswapV2ExchangeAdapterV2.sol
```diff
@@ -102,6 +103,24 @@ contract UniswapV2ExchangeAdapterV2 {
...
+     * @param _fixIn                Boolean representing if input tokens amount is fixed
+     * 
+     * @return bytes                Data parameter to be passed to `getTradeCallData`          
+     */
+    function generateDataParam(address _sellComponent, address _buyComponent, bool _fixIn)
+        external
+        view
+        returns (bytes memory) 
+    {
+        address[2] memory path = [_sellComponent, _buyComponent];
```
bweick, line 120: This is going to need to be a dynamic array type since that's what we'll be expecting when we parse the data. You'll notice we use:
```
        (
            address[] memory path,
            bool shouldSwapTokensForExactTokens
        ) = abi.decode(_data, (address[],bool));
```

Which is expecting to be decoding to a dynamic type.

So this `path` code needs to look something like:
```
address[] memory path = new address[](2);
path[0] = _sellComponent;
path[1] = _buyComponent;
```

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -0,0 +1,585 @@
...
+        address _executor,
+        uint256 _amountSold,
+        uint256 _amountBought,
+        uint256 protcolFee
+    );
+    
+    /* ============ Constants ============ */
+    
+    uint256 private constant TRADE_MODULE_PROTOCOL_FEE_INDEX = 0;
+    uint256 private constant TARGET_RAISE_DIVISOR = 1.0025e18;       // Raise targets 25 bps
```
bweick, line 119: We don't need this anymore since it's state that's set on a per Set basis

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -0,0 +1,699 @@
...
+        uint256 preTradeSendTokenBalance;       // Total initial balance of token being sold
+        uint256 preTradeReceiveTokenBalance;    // Total initial balance of token being bought
+    }
+
+    /* ============ Events ============ */
+
+    event TargetUnitsUpdated(ISetToken indexed _setToken, address indexed _component, uint256 _newUnit, uint256 _positionMultiplier);
+    event TradeMaximumUpdated(ISetToken indexed _setToken, address indexed _component, uint256 _newMaximum);
+    event AssetExchangeUpdated(ISetToken indexed _setToken, address indexed _component, string _newExchangeName);
+    event CoolOffPeriodUpdated(ISetToken indexed _setToken, address indexed _component, uint256 _newCoolOffPeriod);
```
bweick, line 102: Can we add an event for when the raiseTargetPercentage is updated?

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -0,0 +1,699 @@
...
+        _executeTrade(tradeInfo);
+        
+        uint256 protocolFee = _accrueProtocolFee(tradeInfo);
+        
+        (uint256 sellAmount, uint256 buyAmount) = _updatePositionState(tradeInfo);
+
+        executionInfo[_setToken][_component].lastTradeTimestamp = block.timestamp;
+        
+        // todo: should this required be here?
+        // require(buyAmount < executionInfo[_setToken][_component].maxSize, "Trade amount exceeds max allowed trade size");   // should we revert earlier
```
bweick, line 215: Don't need this for now. May add some slippage check which would be done after the trade

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -0,0 +1,699 @@
...
+    event TraderStatusUpdated(ISetToken indexed _setToken, address indexed _trader, bool _status);
+    
+    event TradeExecuted(
+        ISetToken indexed _setToken,
+        address indexed _sellComponent,
+        address indexed _buyComponent,
+        IExchangeAdapter _exchangeAdapter,
+        address _executor,
+        uint256 _amountSold,
+        uint256 _amountBought,
```
bweick, line 114: Can we add a note about whether this amount should be the total buy amount or the amount net of fees just so it's clear

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -0,0 +1,699 @@
...
+
+        _validateTradeParameters(_setToken, _component);
+
+        TradeInfo memory tradeInfo = _createTradeInfo(_setToken, _component);
+        
+        _executeTrade(tradeInfo);
+        
+        uint256 protocolFee = _accrueProtocolFee(tradeInfo);
+        
+        (uint256 sellAmount, uint256 buyAmount) = _updatePositionState(tradeInfo);
```
bweick, line 210: Just as a mental note and to confirm understanding, since we accrue fees before calculating the buyAmount in `_updatePositionState` this `buyAmount` is really net of fees

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -0,0 +1,699 @@
...
+     * exceed components maxTradeSize nor overshoot the target unit. To be used near the end of rebalances when a
+     * component's calculated trade size is greater in value than remaining WETH.
+     *
+     * @param _setToken             Address of the SetToken
+     * @param _component            Address of the SetToken component to trade
+     */
+    function tradeRemainingWETH(ISetToken _setToken, IERC20 _component) external nonReentrant onlyAllowedTrader(_setToken, msg.sender) onlyEOA() virtual {
+
+        require(_noTokensToSell(_setToken), "Sell other set components first");
+        require(executionInfo[_setToken][weth].targetUnit < uint256(_setToken.getDefaultPositionRealUnit(address(weth))), "WETH is below target unit and can not be traded");    // casting?
```
bweick, line 241: We like to use `toUint256()` for casting purposes. Does it safely.

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -0,0 +1,699 @@
...
+        
+        TradeInfo memory tradeInfo = _createTradeInfo(_setToken, _component);
+        
+        _executeTrade(tradeInfo);
+        
+        uint256 protocolFee = _accrueProtocolFee(tradeInfo);
+        
+        (uint256 sellAmount, uint256 buyAmount) = _updatePositionState(tradeInfo);
+
+        require(buyAmount < executionInfo[_setToken][_component].maxSize, "Trade amount exceeds max allowed trade size");   // should we revert earlier
```
bweick, line 253: Technically we need this to `buyAmount.add(protocolFee)` because the `buyAmount` returned here is net of fees

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -0,0 +1,699 @@
...
+
+    constructor(IController _controller, IWETH _weth) public ModuleBase(_controller) {
+        weth = _weth;
+    }
+
+    /* ============ External Functions ============ */
+
+    /**
+     * MANAGER ONLY: Set new target units, zeroing out any units for components being removed from index. Log position multiplier to
+     * adjust target units in case fees are accrued. Validate that all components in current allocation are in _components array.
```
bweick, line 151: Think a better wording for `Validate that all components in current allocation are in _components array.` is `Validate that every oldComponent has a targetUnit`.

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -0,0 +1,692 @@
...
+    /**
+     * ACCESS LIMITED: For situation where all target units met and remaining WETH, uniformly raise targets by same
+     * percentage in order to allow further trading. Can be called multiple times if necessary, increase should be
+     * small in order to reduce tracking error.
+     *
+     * @param _setToken             Address of the SetToken
+     */
+    function raiseAssetTargets(ISetToken _setToken) external onlyManagerAndValidSet(_setToken) virtual {
+        require(
+            _allTargetsMet(_setToken) && _setToken.getDefaultPositionRealUnit(address(weth)) > 0,
```
bweick, line 275: This needs to be greater than the `targetUnit` for `weth`

bweick: And we should make sure we have a test case where we are testing this and the Set has `weth` in it. That goes for all of these functions, we need to make sure it plays well with `weth`.

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -0,0 +1,692 @@
...
+    }
+
+    /**
+     * ACCESS LIMITED: For situation where all target units met and remaining WETH, uniformly raise targets by same
+     * percentage in order to allow further trading. Can be called multiple times if necessary, increase should be
+     * small in order to reduce tracking error.
+     *
+     * @param _setToken             Address of the SetToken
+     */
+    function raiseAssetTargets(ISetToken _setToken) external onlyManagerAndValidSet(_setToken) virtual {
```
bweick, line 273: Let's actually make this callable by `onlyAllowedTrader` instead of only the manager since the manager is parameterizing it by setting the `raiseTargetPercentage` I think we can allow traders to call it

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -0,0 +1,692 @@
...
+     * @param _setToken                     Address of the SetToken
+     * @param _raiseTargetPercentage        Amount to raise all component's unit targets by (in precise units)     
+     */
+    function updateRaiseTargetPercentage(ISetToken _setToken, uint256 _raiseTargetPercentage) external onlyManagerAndValidSet(_setToken) {
+        require(_raiseTargetPercentage > 0, "raiseTargetPercentage > 0");
+        rebalanceInfo[_setToken].raiseTargetPercentage = _raiseTargetPercentage;        
+    }
+    
+    /**
+     * MANAGER ONLY: Toggle ability for passed addresses to trade from current state
```
bweick, line 367: Can remove `from current state`

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -0,0 +1,692 @@
...
+            executionInfo[_setToken][IERC20(position.component)].lastTradeTimestamp = 0;
+        }
+
+        _setToken.initializeModule();
+    }
+    
+    function removeModule() external override {}
+    
+    // function removeModule() external override {
+    //     delete executionInfo[ISetToken(msg.sender)];    // todo: figure out how to delete efficiently
```
bweick, line 419: This may just require cycling through everything. It'll be worth it I think cause you'll be zeroing out state. We can also ask our auditors for recommendations so I would struggle too much over this

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -0,0 +1,692 @@
...
+     *
+     * @return TradeInfo            Struct containing data for trade
+     */
+    function _createTradeRemainingInfo(ISetToken _setToken, IERC20 _component) internal view returns (TradeInfo memory) {
+        
+        uint256 totalSupply = _setToken.totalSupply();
+
+        uint256 componentMaxSize = executionInfo[_setToken][_component].maxSize;
+        
+        uint256 currentUnit = _setToken.getDefaultPositionRealUnit(address(_component)).toUint256();
```
bweick, line 503: We care about `weth` size here not the component

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -0,0 +1,692 @@
...
+        uint256 targetUnit = _normalizedTargetUnit(_setToken, _component);
```
bweick, line 504: Ditto to above

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -0,0 +1,692 @@
...
+     * @return bool                 True if there is not any component that can be sold, otherwise false
+     */
+    function _noTokensToSell(ISetToken _setToken) internal view returns (bool) {
+        uint256 positionMultiplier = rebalanceInfo[_setToken].positionMultiplier;
+        uint256 currentPositionMultiplier = _setToken.positionMultiplier().toUint256();
+        address[] memory rebalanceComponents = rebalanceInfo[_setToken].rebalanceComponents;
+        for (uint256 i = 0; i < rebalanceComponents.length; i++) {
+            address component = rebalanceComponents[i];
+            if (component != address(weth)) {
+                uint256 normalizedTargetUnit = executionInfo[_setToken][IERC20(component)].targetUnit.mul(currentPositionMultiplier).div(positionMultiplier);
```
bweick, line 624: Think we can use `_normalizedTargetUnit` here

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -0,0 +1,692 @@
...
+     * @return bool                 True if all component's target units have been met, otherwise false
+     */
+    function _allTargetsMet(ISetToken _setToken) internal view returns (bool) {
+        uint256 positionMultiplier = rebalanceInfo[_setToken].positionMultiplier;
+        uint256 currentPositionMultiplier = _setToken.positionMultiplier().toUint256();
+        address[] memory rebalanceComponents = rebalanceInfo[_setToken].rebalanceComponents;
+        for (uint256 i = 0; i < rebalanceComponents.length; i++) {
+            address component = rebalanceComponents[i];
+            if (component != address(weth)) {
+                uint256 normalizedTargetUnit = executionInfo[_setToken][IERC20(component)].targetUnit.mul(currentPositionMultiplier).div(positionMultiplier);
```
bweick, line 646: Think we can use `_normalizedTargetUnit`

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -0,0 +1,692 @@
...
+    }
+
+    /**
+     * Check if all targets are met
+     *
+     * @param _setToken             Instance of the SetToken to be rebalanced
+     *
+     * @return bool                 True if all component's target units have been met, otherwise false
+     */
+    function _allTargetsMet(ISetToken _setToken) internal view returns (bool) {
```
bweick, line 639: IMPORTANT: When testing for this function we need to do a case where the Set's `positionMultiplier` is not equal to `10^18`. So may want to accrue fees.

contracts/protocol/modules/GeneralIndexModule.sol
Alpha-Guy: Yes.

contracts/protocol/modules/GeneralIndexModule.sol
Alpha-Guy: The issue was `_normalizedTargetUnit` was loading variables `currentPositionMultiplier` and `positionMultiplier` to perform the caluculation. If we call that function in a for loop, it would load the variables in memory, multiple times. Hence I was not calling the function.

But now, I have separated the `_normalizedTargetUnit` function to 
- `_getNormalizedTargetUnit` - loads the variable and calls the `_normalizedTargetUnit`
- `_normalizedTargetUnit` - receives all the necessary values as parameter and only performs calculation
And I am calling `_noramalizedTargetUnit` within the for loop.

bweick: Sweet that makes a lot of sense

test/protocol/modules/generalIndexModule.spec.ts
```diff
@@ -0,0 +1,2080 @@
...
+        newTargetUnits = [];
+        oldTargetUnits = [ether(100), ZERO, ether(175)];  // 10$ of WETH should be extra in the SetToken
+        issueAmount = ether("20.000000000000000000");
+        subjectRaiseTargetPercentage = ether("0.02");
+
+        subjectSetToken = index;
+        subjectIncreaseTime = ONE_MINUTE_IN_SECONDS.mul(5);
+      });
+
+      describe("#updateRaiseTargetPercentage", async () => {
```
bweick, line 1803: This should be in its own describe block, not a sub block of raiseAssetTargets. If they share some context you can use a context block. But each tested function should start at the same level.

test/protocol/modules/generalIndexModule.spec.ts
```diff
@@ -0,0 +1,2080 @@
...
+        before(async () => {
+          subjectCaller = trader;
+        });
+
+        beforeEach(async () => {
+          await setup.approveAndIssueSetToken(subjectSetToken, issueAmount);
+          await indexModule.startRebalance(subjectSetToken.address, newComponents, newTargetUnits, oldTargetUnits, await index.positionMultiplier());
+        });
+
+        async function subject(): Promise<ContractTransaction> {
```
bweick, line 1851: The subject should simply be calling the function. With this set up we can't isolate what went wrong. Was it the raiseTargets? Or was it one of the trades reverting? Also we don't need to make the trades after raising the targets, all we need to do is check that the new targets are what we expect.

test/protocol/modules/generalIndexModule.spec.ts
```diff
@@ -0,0 +1,2080 @@
...
+
+          // trade WETH for dai, to reach dai target unit
+          await increaseTimeAsync(ONE_MINUTE_IN_SECONDS.mul(5));
+          await indexModule.connect(trader.wallet).trade(subjectSetToken.address, setup.dai.address, MAX_UINT_256);
+          await increaseTimeAsync(ONE_MINUTE_IN_SECONDS.mul(5));
+          await indexModule.connect(trader.wallet).trade(subjectSetToken.address, setup.dai.address, MAX_UINT_256);
+        }
+        it("should raise asset targets and allow trading", async () => {
+
+          await indexModule.connect(owner.wallet).updateRaiseTargetPercentage(subjectSetToken.address, subjectRaiseTargetPercentage);
```
bweick, line 1887: Everything from line 1887-1893 should be in a before or beforeEach block. We want the actual block where we test the case to only contain the `subject` and and calls we need to make on-chain to check state. There should be no other transactions.

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -0,0 +1,721 @@
...
+
+        address[] memory aggregateComponents = currentComponents.extend(_newComponents);
+        uint256[] memory aggregateTargetUnits = _oldComponentsTargetUnits.extend(_newComponentsTargetUnits);
+
+        require(!aggregateComponents.hasDuplicate(), "Cannot duplicate components");
+
+        for (uint256 i = 0; i < aggregateComponents.length; i++) {
+            
+            executionInfo[_setToken][IERC20(aggregateComponents[i])].targetUnit = aggregateTargetUnits[i];
+
```
bweick, line 187: You had a note somewhere about validating that each component has a specified exchange. That's a great catch and I think we can check that here when we cycle through the components

test/protocol/modules/generalIndexModule.spec.ts
```diff
@@ -0,0 +1,2080 @@
...
+
+          await subject();
+
+          const wethPositionUnits = await subjectSetToken.getDefaultPositionRealUnit(setup.weth.address);
+          const daiPositionUnits = await subjectSetToken.getDefaultPositionRealUnit(setup.dai.address);
+
+          const expectedWethPositionUnits = preciseDiv(wethAmount.sub(remainingWETH), totalSupply);
+          const expectedDaiPositionUnits = preciseDiv(daiAmount.add(expectedSubjectAmountOut), totalSupply);
+
+          expect(wethPositionUnits).to.eq(expectedWethPositionUnits);
```
bweick, line 1702: It'd be nice to have an explicit check that the `WETH` units end at the targetUnit set by the manager.

test/protocol/modules/generalIndexModule.spec.ts
```diff
@@ -0,0 +1,2080 @@
...
+          exchanges = [uniswapAdapterName, sushiswapAdapterName];
+        });
+
+        it("should revert", async () => {
+          await expect(subject()).to.be.revertedWith("Array length must be > 0");
+        });
+      });
+    });
+
+    describe("#trade", async () => {
```
bweick, line 566: One thing that's missing here is a check to make sure the transaction reverts if the exchange would return less than or take more than the amount we want. This is where mocking the exchanges is nice, because we could actually check what is passed into our mocks, we can't do that with these black box exchanges we deployed on our local environments.

In this case let's calculate the expectedOut or In and then manipulate the `ethQuantityLimit` to make it revert as we would expect.

test/protocol/modules/generalIndexModule.spec.ts
```diff
@@ -0,0 +1,2080 @@
...
+          setup.weth.address,
+          ether(1000),
+          THREE
+        )).totalOutput;
+        ethQuantityLimit = expectedOut;
+      });
+
+      async function subject(): Promise<ContractTransaction> {
+        await increaseTimeAsync(subjectIncreaseTime);
+        return await indexModule.connect(subjectCaller.wallet).trade(subjectSetToken.address, subjectComponent, ethQuantityLimit);
```
bweick, line 599: This should be `subjectEthQuantityLimit`. Always prefix subject inputs with subject so it's clear what's being entered into the tests.

test/protocol/modules/generalIndexModule.spec.ts
```diff
@@ -0,0 +1,2080 @@
...
+          await expect(subject()).to.be.revertedWith("Array length must be > 0");
+        });
+      });
+    });
+
+    describe("#trade", async () => {
+      let subjectComponent: Address;
+      let subjectIncreaseTime: BigNumber;
+      let expectedOut: BigNumber;
+      let ethQuantityLimit: BigNumber;
```
bweick, line 570: `subjectEthQuantityLimit`, and it's a small thing but lets group all the "subject" params together

test/protocol/modules/generalIndexModule.spec.ts
```diff
@@ -0,0 +1,2080 @@
...
+          remainingWETH.gt(ZERO) ? remainingWETH : ZERO,
+          THREE
+        )).totalOutput;
+      });
+
+      async function subject(): Promise<ContractTransaction> {
+        return await indexModule.connect(subjectCaller.wallet).tradeRemainingWETH(
+          subjectSetToken.address,
+          subjectComponent,
+          expectedSubjectAmountOut
```
bweick, line 1579: `subjectComponentQuantityLimit`! Let's also add a test to make sure it will revert if we don't get enough component back.

test/protocol/modules/generalIndexModule.spec.ts
```diff
@@ -0,0 +1,2080 @@
...
+      //     await indexModule.setExchanges(subjectSetToken.address, [subjectComponent], [""]);
+      //   });
+      //   it("the trade reverts", async () => {
+      //     await expect(subject()).to.be.revertedWith("Exchange must be specified");
+      //   });
+      // });
+    });
+
+    describe("#tradeRemainingWETH", async () => {
+      let subjectComponent: Address;
```
bweick, line 1539: Group the subject variable together, easier to understand the test set up

test/protocol/modules/generalIndexModule.spec.ts
```diff
@@ -0,0 +1,2080 @@
...
+        });
+      });
+    });
+
+    describe("#raiseAssetTargets", async () => {
+      let subjectIncreaseTime: BigNumber;
+      let subjectSetToken: SetToken;
+      let subjectRaiseTargetPercentage: BigNumber;
+
+      before(async () => {
```
bweick, line 1792: Small nit: but to make it easier to review let's try to be consistent about where variables are declared and placed. In general, unless absolutely necessary, we want all the "subject" variables declared at the end of the block and grouped together. Makes it much easier to review code that way:

```
      before(async () => {
        newComponents = [];
        newTargetUnits = [];
        oldTargetUnits = [ether(100), ZERO, ether(175)];  // 10$ of WETH should be extra in the SetToken
        issueAmount = ether("20.000000000000000000");

        subjectRaiseTargetPercentage = ether("0.02");
        subjectSetToken = index;
        subjectIncreaseTime = ONE_MINUTE_IN_SECONDS.mul(5);
      });
      ```

File: test/protocol/modules/generalIndexModule.spec.ts [hunk 30]
```diff
@@ -0,0 +1,2080 @@
...
+        });
+
+        it("should revert", async () => {
+          await expect(subject()).to.be.revertedWith("Address not permitted to trade");
+        });
+      });
+
+      describe("when set has weth as component", async () => {
+        before(async () => {
+          subjectSetToken = indexWithWeth;
```
bweick, line 1682: Declare at end of block and group together subject variables

contracts/protocol/modules/GeneralIndexModule.sol
Alpha-Guy: Should we allow an empty string for `weth`? Or should `weth` be removed from the components list when setting exchanges.

bweick: Hmm great point I would just skip the check if the component is WETH

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -419,30 +462,58 @@ contract GeneralIndexModule is ModuleBase, ReentrancyGuard {
...
         _setToken.initializeModule();
     }
     
-    function removeModule() external override {}
+    /**
+     * Called by a SetToken to notify that this module was removed from the SetToken. 
+     * Clears the state of the calling SetToken.
+     */
+    function removeModule() external override {
+        // delete executionInfo[ISetToken(msg.sender)];    // todo: figure out how to delete efficiently
```
bweick, line 475: Remove this line

contracts/protocol/integration/BalancerV1ExchangeAdapter.sol
```diff
@@ -0,0 +1,128 @@
...
+        external
+        view
+        returns (address, uint256, bytes memory)
+    {   
+        (
+            bool shouldSwapFixedAmount
+        ) = abi.decode(_data, (bool));
+
+        bytes memory callData = abi.encodeWithSignature(
+            shouldSwapFixedAmount ? EXACT_IN : EXACT_OUT,
```
bweick, line 92: Lets make this naming clearer. I can't glean from this name which side of the trade should be fixed

contracts/protocol/integration/BalancerV1ExchangeAdapter.sol
```diff
@@ -0,0 +1,128 @@
...
+     * @return address                   Target contract address
+     * @return uint256                   Call value
+     * @return bytes                     Trade calldata
+     */
+    function getTradeCalldata(
+        address _sourceToken,
+        address _destinationToken,
+        address _destinationAddress,
+        uint256 _sourceQuantity,
+        uint256 _minDestinationQuantity,
```
bweick, line 80: Let's change this to `_destinationQuantity` like it is for the Uniswap adapter. Make sure to update the javadocs to reflect that (as well as the `Note`)

contracts/protocol/integration/BalancerV1ExchangeAdapter.sol
```diff
@@ -0,0 +1,128 @@
...
+            shouldSwapFixedAmount ?  _sourceQuantity : _minDestinationQuantity,
+            shouldSwapFixedAmount ? _minDestinationQuantity : _sourceQuantity,
+            BALANCER_POOL_LIMIT
+        );
+
+        return (balancerProxy, 0, callData);
+    }
+
+    /**
+     * Generate data parameter to be passed to `getTradeCallData`. Returns encoded trade paths and bool to select trade function.
```
bweick, line 104: Update javadoc, doesn't return a path for Balancer

contracts/protocol/integration/BalancerV1ExchangeAdapter.sol
```diff
@@ -0,0 +1,128 @@
...
+
+pragma solidity 0.6.10;
+pragma experimental "ABIEncoderV2";
+
+/**
+ * @title BalancerV1ExchangeAdapter
+ * @author Set Protocol
+ *
+ * A Balancer exchange adapter that returns calldata for trading.
+ *
```
bweick, line 27: Remove this extra line

contracts/protocol/integration/UniswapV2ExchangeAdapterV2.sol
```diff
@@ -92,16 +93,36 @@ contract UniswapV2ExchangeAdapterV2 {
...
+     */
+    function generateDataParam(address _sellComponent, address _buyComponent, bool _fixIn)
+        external
+        view
+        returns (bytes memory) 
+    {
+        address[] memory path = new address[](2);
+        path[0] = _sellComponent;
+        path[1] = _buyComponent;
+        bool shouldSwapTokensForExactTokens = _fixIn ? false : true;
```
bweick, line 122: Should be able to just make this:
`bool shouldSwapTokensForExactTokens = !_fixIn;`

bweick: Or even better just put that right into the encoding function

bweick: If you put it into the `encode` function make sure to add a comment that the returned boolean is for a fixed amount OUT

contracts/protocol/integration/BalancerV1ExchangeAdapter.sol
bweick, line 92: Let's choose a pattern between the Uniswap and Balancer adapters. Is the passed boolean to fix the input amount or the output amount? It get's too confusing otherwise.

bweick: For Balancer the boolean passed in the data indicates whether the amount_in is fixed. For Uniswap the boolean indicated whether the amount_out is fixed.

contracts/protocol/integration/BalancerV1ExchangeAdapter.sol
```diff
@@ -128,7 +127,7 @@ contract BalancerV1ExchangeAdapter {
      *
      * @return bytes               Encoded data used for trading on Balancer
      */
-    function getBalancerExchangeData(bool _shouldSwapForFixedOutputAmount) external view returns (bytes memory) {
-        return abi.encode(_shouldSwapForFixedOutputAmount);
+    function getBalancerExchangeData(bool _shouldSwarFixeInputtAmount) external view returns (bytes memory) {
```
bweick, line 130: variable naming typo here

bweick (approved):

PR #64, opened 2021-04-09
Add forked provider infrastructure and simple TradeModule / ExchangeAdapter tests

cgewecke:
Adding some tests that validate logic against forked mainnet was recently recommended by an auditor. These are intended to be supplementary to tests with full mocks (and don't contribute to measured test coverage).

They are useful for:
+ testing operational edge-cases like sequences of trades
+ sanity checking the data generated by method signature encoding

PR:
+ Upgrades Hardhat to latest (2.1.2) (the forking feature receives regular bug-fixes)
+ Adds a forked provider job in CI (and cleans up the CI script a bit)
+ Adds forked integration tests for simple trades for ZeroEx, Uniswap, and SushiSwap.

Leaving comments in the diff below about misc. design issues...

~~(Unfortunately this is WIP because I can't get the ZeroEx tests to work as expected...will circle back here.)~~


.env.default
```diff
@@ -1,4 +1,5 @@
 # These are randomly generated hex so that CircleCI will work
+ALCHEMY_TOKEN=fake_alchemy_token
```
cgewecke, line 2: Adding an Alchemy token to the credential requirements. (Have added one to the CI environment variables)

This is necessary because Infura doesn't provide access to archive node data and the forked tests may request it.

contracts/mocks/ForceFunderMock.sol
```diff
@@ -0,0 +1,33 @@
...
+     * Convenience method for depositing eth into non-payable contracts
+     * which the forked provider tests would like to impersonate
+     * as a message sender.
+     *
+     * @param  destination   destination of eth payment
+     */
+    function fund(address destination) public payable {
+        selfdestruct(payable(address(destination)));
+    }
+}
```
cgewecke, line 33: This contract a trick to force an ether balance into a non-payable contract using self destruct. This is useful when you want to impersonate (send a message from) a non-payable forked contract and it has to have an eth balance to fund the tx.

utils/deploys/dependencies.ts
```diff
@@ -167,6 +170,14 @@ export default {
...
     50: "test-rpc",
   },
+
+  // WHALES (for forked mainnet testing)
+
+  USDC_WHALE: "0x47ac0Fb4F2D84898e4D9E7b4DaB3C24507a6D503",
+  DAI_WHALE: "0x47ac0Fb4F2D84898e4D9E7b4DaB3C24507a6D503",
+  WETH_WHALE: "0x94B0A3d511b6EcDb17eBF877278Ab030acb0A878",
+  WBTC_WHALE: "0x2260fac5e5542a773aa44fbcfedf7c193bc2c599",
+
```
cgewecke, line 180: These are really contracts, not whales. They're used as a source of tokens to transfer into the Set system.

utils/test/accountUtils.ts
```diff
@@ -39,3 +43,52 @@ export const getEthBalance = async (account: Address): Promise<BigNumber> => {
...
+  for (const whale of whales) {
+    await network.provider.request({
+      method: "hardhat_impersonateAccount",
+      params: [whale]},
+    );
+
+    const funder = await deployer.mocks.deployForceFunderMock();
+    await funder.fund(whale, {value: ether(100)}); // Gas money
+  }
+};
```
cgewecke, line 93: This method initializes the connected signers of the forked token instances so we can call `transfer` on the token from a hardhat "impersonated", gas funded, mainnet token holder with adequate reserves.

cgewecke:
Am removing the skipped ZeroEx adapter test because fails with an internal hardhat node error:
```
TransactionExecutionError: Number can only safely store up to 53 bits      
    at HardhatNode.mineBlock (...hardhat-network/provider/node.ts:327:13)
```
Initially thought this was a hardhat bug or error in my test setup but was running forked provider against a ZeroEx swap endpoint for `set.js` this week and saw the same error *intermittently* ... now I think this is likely something about the way 0x interacts with forking and that it may not be possible to get quotes for a block in the past. 

The forked block is pinned in these tests for reproducibility.

bweick (approved):

PR #75, opened 2021-04-19
GeneralIndexModule internal review and external audit revisions

cgewecke:
This is a set of small revisions suggested by the internal review of GeneralIndexModule. Each idea is expressed as a commit, so it can be considered on it's own and cherry-picked if useful. 

The existing contract is highly optimized for gas - it minimizes state reads in loops as well as method encapsulation that would impose costs for extra function calls. 

This version of the contract does the opposite - it introduces additional getters and generally ignores state read costs. For example, `_normalizedTargetUnits` is removed. It's also more aggressive about param validation. checks are added for 
+ the public getters: these might be used in automated scripts and ambiguous outputs are generated for invalid inputs
+ the exchange name is validated in `setExchange` (it's also validated during trades). Also storing this as a bytes32 hash, mirroring the IntegrationRegistry.  

The Berlin hardfork has made initial state reads for a location more expensive, but subsequent "warm" reads from the same place are much cheaper, changing the gas calculus a bit. Despite the inefficiencies introduced in this draft, the diff between versions suggests that gas costs increased only marginally overall. They would increase more in a module with a larger number of components.

-----

#### Update I

Additional commits have been added after the review was finalized on Monday. Notable things are:
+ storage of exchangeNames as bytes32 hashes has been reverted following initial review here - these are strings again
+ *initialize* and *startRebalance* now revert if there are external positions for a component
+ *_createTradeInfo* and *_createTradeRemainingInfo* have been substantially restructured 
  + common logic has been pulled out and encapsulated in additional helpers
  + the field naming for the TradeInfo struct has changed to more closely match TradeModule's syntax: 
    + from "fixed/floatingQuantity" to `sendQuantity` and `minReceiveQuantity`
    + from `isSendFixed` to `isSell`  

------

**Update II:** (Internal review remainders, review feedback, audit corrections)

TODO:
- [x] Revert `isSell`, `minReceivedQuantity` to `isFixedSend`, `floatingQuantityLimit`
- [x] Move _validate array methods to AddressArrayLibs (and add additional)
- [x] Optimize external position detection
- [x] Optimize modifier inlining
- [x] Declare TradeInfo as return parameter
- [x] Add `Position.unitsMatchWithinRange` method for targets logic (invert)
- [x] Restrict `_getAggregateComponentsAndUnits` state mutability to `pure`
- [x] Iosoro 5.3.2: add `removeModule` tests for preserved state, eval trader deletion
- [x] Iosoro 5.4.1: event (w/ position multiplier) for `raiseAssetTarget`
- [x] Iosoro 5.4.3: immutable weth
- [x] Iosoco 5.4.3: typos
- [x]  Ask Felix for clarification about fees ambiguity in *[_calculateTradeSizeAndDirection][22]*
  + Response:
  + > It’s a code organization comment, where either 1) all fee-related calculations and logic should be grouped together or 2) a big comment is used to flag that logic and how it interacts w/ the rest of the module. In addition, I wanted to ascertain that there are no problematic edge cases and that the math is correct
- [x]  Update gas analysis
- [x] Re-review docs and normalize namings
- [x]  Address fee concern on *_calculateTradeSizeAndDirection*
  + move fn level comment about issue inline
  + check test cases
- [x] fix validatePairs bytes array method
- [x] rename PositionMultiplier update --> AssetTargetsRaised
- [x] remove de-authorized traders from history
- [x] add changelog entry in ... (one of the libs)

[22]: https://github.com/SetProtocol/set-protocol-v2/blob/0a86d8777153fd6410577b5bd995542927e7b8b3/contracts/protocol/modules/GeneralIndexModule.sol#L771-L787
Report shows original GeneralIndexModule vs. this PR (run on hardhat@2.2.0-rc.2)

(Updated for most recent commits - April 20)

## Deployments
|                                |       Gas |                                                                      |            Diff | Diff % | Block % | usd avg |
| :----------------------------- | --------: | :------------------------------------------------------------------: | --------------: | -----: | ------: | ------: |
| **GeneralIndexModule**         | 4,358,869 | passed | [**-45,230**]() |    -1% |    35 % | 1389.93 |

## Methods
|                                       |       Gas |                                                                      |            Diff | Diff % | Calls | usd avg |
| :------------------------------------ | --------: | :------------------------------------------------------------------: | --------------: | -----: | ----: | ------: |
| **GeneralIndexModule**                |           |                                                                      |                 |        |       |         |
|        *initialize*                   |   189,082 | passed | [**-35,895**]() |   -16% |    12 |   60.29 |
|        *raiseAssetTargets*            |    82,425 | failed |  [**+1,159**]() |    +1% |     6 |   26.28 |
|        *setAnyoneTrade*               |    62,487 |                                                                      |               - |      - |     4 |   19.93 |
|        *setCoolOffPeriods*            |   121,277 | failed |  [**+7,130**]() |    +6% |     4 |   38.67 |
|        *setExchanges*                 |   120,933 | failed | [**+31,341**]() |   +35% |     6 |   38.56 |
|        *setRaiseTargetPercentage*     |    62,616 |                                                                      |               - |      - |    12 |   19.97 |
|        *setTradeMaximums*             |    93,017 | failed |  [**+8,397**]() |   +10% |     7 |   29.66 |
|        *setTraderStatus*              |   109,688 |                                                                      |               - |      - |    13 |   34.98 |
|        *startRebalance*               |   193,078 | failed | [**+23,031**]() |   +14% |    30 |   61.57 |
|        *trade*                        |   364,066 | passed |  [**-1,932**]() |    -1% |    59 |  116.09 |
|        *tradeRemainingWETH*           |   286,393 | passed |    [**-226**]() |     0% |     9 |   91.32 |

## Build Configuration
| Option                 | Settings        |
| ---------------------- | --------------- |
| solc: version          | 0.6.10          |
| solc: optimized        | true            |
| solc: runs             | 200             |
| gas: block limit       | 12,450,000      |
| gas: price             | 123 gwei/gas    |
| gas: currency/eth rate | 2592.47 usd/eth |


bweick:
What are the benefits of storing the exchange name as a hash? To me this hurts readability.

bweick:
So with Berlin would we expect the gas increase to be slightly less? (I'm assuming this wasn't run with Berlin changes)

cgewecke:
> What are the benefits of storing the exchange name as a hash? To me this hurts readability.

Agree, it's not as nice. There's a note in the review about the cost of storing them as a string.

> Note that exchangeNames as strings will be very expensive to store and update

cgewecke:
@bweick Yes, this was run on a Berlin EVM. In Istanbul I'd expect these changes to increase costs a lot, but with Berlin they're more expensive but...by a smaller margin.

bweick:
> > What are the benefits of storing the exchange name as a hash? To me this hurts readability.
> 
> Agree, it's not as nice. There's a note in the review about the cost of storing them as a string.
> 
> > Note that exchangeNames as strings will be very expensive to store and update

Gotcha, that makes sense, though am I reading the table right that the cost of setting exchanges went up?

cgewecke:
@bweick Yes, it went up quite a bit, but it's also validating the exchange name now. From the internal review....

> Validate that the exchangeName is on the integrationRegistry

This is a redundant check b/c the trade methods *also* check. But maybe makes sense from a UX perspective? In practice how frequently are the settings changed?

bweick:
> @bweick Yes, it went up quite a bit, but it's also validating the exchange name now. From the internal review....
> 
> > Validate that the exchangeName is on the integrationRegistry
> 
> This is a redundant check b/c the trade methods _also_ check. But maybe makes sense from a UX perspective? In practice how frequently are the settings changed?

I like the check here, if we could remove the check on the trade that'd be great since that is redundant but I think our inherited helper functions there enforce a check?

Maybe we can discuss a bit as a group but the readability is really nice. Do we know what the marginal cost of storing a string vs bytes32 is? I know it's going to be different based on string length but would be great to isolate the "cost of readability".

cgewecke:
@bweick 

Changing back to string increases the costs for `setExchanges` 14% (from this PR as a baseline, with the tests using 4 components). The deployment cost increased 3%.

With a single exchange name longer than 32bytes `setExchanges` increased 24%. (Ex: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa)

There was no impact on the trade methods.

In the end this is just a few dollars for sets with a small number of components...if exchanges aren't configured very often idk that these gas differences matter that much.

cgewecke:
> if we could remove the check on the trade that'd be great

Do you know if this is possible?  I was thinking that validating the adapter is necessary in trade because it could have been removed at some point between setting the exchanges and calling trade.

bweick:
> > if we could remove the check on the trade that'd be great
> 
> Do you know if this is possible? I was thinking that validating the adapter is necessary in trade because it could have been removed at some point between setting the exchanges and calling trade.

Well the only way it could be removed is through setExchanges, in which case a valid exchange would need to be passed in its place. Unless there's a path to removal I'm not thinking of?

cgewecke:
Can the adapter be removed via the IntegrationRegistry contract? I wrote a test for this possibility in #trade but maybe it's not right...

https://github.com/SetProtocol/set-protocol-v2/blob/6f96431d636915070e7deff6760220303bc1b4b8/test/protocol/modules/generalIndexModule.spec.ts#L910-L927

bweick:
My one gripe with this `from "fixed/floatingQuantity" to sendQuantity and minReceiveQuantity` is that `minReceiveQuantity` is a misnomer since it can also represent `maxSpendQuantity` which is why I named it something more generic like floating. Is there a better way to encapsulate that?

Would also note that `isSell` is as bit of a misnomer when it comes to `tradeRemainingWETH` since WETH is the fixed quantity but in general `isSell` feels like it refers to is the component being sold.

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -890,34 +956,134 @@ contract GeneralIndexModule is ModuleBase, ReentrancyGuard {
...
      *
-     * @param _components           Array of components
+     * @param _addresses            Array of addresses
      * @param _data                 Array of uint256 values
      */
-    function _validateUintArrays(address[] calldata _components, uint256[] calldata _data) internal pure {
-        require(_components.length == _data.length, "Array length mismatch");
-        require(_components.length > 0, "Array length must be > 0");
-        require(!_components.hasDuplicate(), "Cannot duplicate components");
+    function _validateAddressAndUintArrays(address[] calldata _addresses, uint256[] calldata _data) internal pure {
```
bweick, line 1010: Wonder if it's worthwhile putting these into a util file?

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -625,51 +620,81 @@ contract GeneralIndexModule is ModuleBase, ReentrancyGuard {
...
+     * @param  calculateTradeDirection      Indicates whether method should calculate trade size and direction
+     *
+     * @return {TradeInfo}                  Struct containing partial data for trade
+     */
+    function _getDefaultTradeInfo(ISetToken _setToken, IERC20 _component, bool calculateTradeDirection)
+        internal
+        view
+        returns (TradeInfo memory)
+    {
         TradeInfo memory tradeInfo;
```
bweick, line 669: Can we just declare this in the `returns` statement? Something like `returns (TradeInfo memory tradeInfo)`

cgewecke:
> My one gripe with this from "fixed/floatingQuantity" to sendQuantity and minReceiveQuantity is that minReceiveQuantity is a misnomer since it can also represent maxSpendQuantity which is why I named it something more generic like floating. Is there a better way to encapsulate that?

I'm not sure...the idea from the review was to remove ternary conditional setting logic in `_executeTrade` and elsewhere and have the TradeInfo constructors generate an object that's in the format consumed by the TradeModule, using its naming. And that. maybe taking that approach simplifies things a bit structurally.  

Is your gripe that the code has lost descriptive clarity with this renaming? If so, maybe a way of addressing that is to add inline comments explaining how floating maps onto this.

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -767,119 +798,154 @@ contract GeneralIndexModule is ModuleBase, ReentrancyGuard {
...
     }
 
     /**
-     * Normalize target unit to current position multiplier in case position multiplier has been changed since startRebalance()
-     * was last called.
+     * Calculates and returns the normalized target unit value.
      *
-     * @param _setToken             Instance of the SetToken to be rebalanced
-     * @param _component            IERC20 component whose normalized target unit is required
+     * @param _setToken                         Instance of the SettToken to be rebalanced
```
cgewecke, line 889: Mis-spelling... and revert space changes

contracts/protocol/modules/GeneralIndexModule.sol
cgewecke: Yes maybe they could be added to `AddressArratUtils` as 
```
pairsWithUintArray
pairsWithBoolArray
...
```
?

bweick:
> a

Yeah maybe linking it all together is some contextual explanation that if isSell is true then the sendToken quantity is a fixed amount. I think that's sufficient we just need to make sure that context is somewhere.

Secondarily, I still think it's weird we call something `minReceiveQuantity` when it could actually be the opposite. I know the idea is to make it match the TradeModule but imo the TradeModule is too simple in that it only allows trading in one direction (fixed amount in), so I'm not sure that's the best thing to anchor our reasoning on.

cgewecke:
>  I still think it's weird we call something minReceiveQuantity when it could actually be the opposite. I know the idea is to make it match the TradeModule but imo the TradeModule is too simple in that it only allows trading in one direction (fixed amount in), so I'm not sure that's the best thing to anchor our reasoning on.

Yes, you're right. This is too weird.

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -321,40 +315,45 @@ contract GeneralIndexModule is ModuleBase, ReentrancyGuard {
...
      * @param _setToken             Address of the SetToken
      * @param _components           Array of components
      * @param _tradeMaximums        Array of trade maximums mapping to correct component
      */
     function setTradeMaximums(
         ISetToken _setToken,
-        address[] calldata _components,
-        uint256[] calldata _tradeMaximums
+        address[] memory _components,
+        uint256[] memory _tradeMaximums
```
cgewecke, line 351: This change from `calldata` to `memory` (in four methods) is due to migrating the array pair validators to AddressArrayUtils. There are compiler errors for both permutations of calldata/memory
+ preserving `calldata` in this file and making the method signatures at lib and mocks `calldata` results in solc erroring with
```
Using solcjs
Compiling 118 files with 0.6.10
InternalCompilerError: Invalid conversion to calldata type.

Error HH600: Compilation failed
```
+ Keeping `calldata` signatures here and `memory` signatures there has solc error with:
```
Using solcjs
Compiling 118 files with 0.6.10
UnimplementedFeatureError: Copying nested dynamic calldata arrays to memory is not implemented in the old code generator.

Error HH600: Compilation failed
```

contracts/lib/PreciseUnitMath.sol
```diff
@@ -187,4 +187,11 @@ library PreciseUnitMath {
...
     }
-}
\ No newline at end of file
+
+    /**
+     * @dev Returns true if a =~ b within range, false otherwise.
+     */
+    function approximatelyEquals(uint256 a, uint256 b, uint256 range) internal pure returns (bool) {
+        return a <= b.add(range) && a >= b.sub(range);
+    }
```
cgewecke, line 196: Note: this returns true if vals are equal up to and including +/- range (preserving the logic at [_targetUnmet][1])

[1]: https://github.com/SetProtocol/set-protocol-v2/blob/0a86d8777153fd6410577b5bd995542927e7b8b3/contracts/protocol/modules/GeneralIndexModule.sol#L836

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -328,6 +329,7 @@ contract GeneralIndexModule is ModuleBase, ReentrancyGuard {
         rebalanceInfo[_setToken].positionMultiplier = rebalanceInfo[_setToken].positionMultiplier.preciseDiv(
             PreciseUnitMath.preciseUnit().add(rebalanceInfo[_setToken].raiseTargetPercentage)
         );
+        emit PositionMultiplierUpdated(_setToken, rebalanceInfo[_setToken].positionMultiplier);
```
bweick, line 336: don't think we need this since it will be emitted on the SetToken

bweick: I think a better event here is just noting that the AssetTargets have been raised. Could maybe log the new positionMultiplier in there

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -450,6 +451,9 @@ contract GeneralIndexModule is ModuleBase, ReentrancyGuard {
         _validateAddressAndBoolArrays(_traders, _statuses);
 
         for (uint256 i = 0; i < _traders.length; i++) {
+            if (_statuses[i]) {
+                permissionInfo[_setToken].tradersHistory.push(_traders[i]);
```
bweick, line 455: Do we care to remove any traders that get set from `true` back to `false`?

cgewecke: 215b6ef

contracts/protocol/modules/GeneralIndexModule.sol
cgewecke: 00f49b8

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -1038,6 +1036,25 @@ contract GeneralIndexModule is ModuleBase, ReentrancyGuard {
...
+     * @param _setToken                         Instance of the SetToken
+     * @param _trader                           Trader whose permission is being set
+     * @param _status                           Boolean permission being set
+
+     */
+    function _updateTradersHistory(ISetToken _setToken, address _trader, bool _status) internal {
+        if (_status) {
+            permissionInfo[_setToken].tradersHistory.push(_trader);
+        } else if(permissionInfo[_setToken].tradersHistory.contains(_trader)) {
+            permissionInfo[_setToken].tradersHistory = permissionInfo[_setToken].tradersHistory.remove(_trader);
```
bweick, line 1056: We have a method in `AddressArrayUtils` that's way cheaper called `removeStorage`, better than having to read it into memory, manipulate, and store.

test/protocol/modules/generalIndexModule.spec.ts
```diff
@@ -2187,6 +2187,24 @@ describe("GeneralIndexModule", () => {
         );
       });
 
+      describe("when de-authorizing a trader", async () => {
```
bweick, line 2190: Probably worth checking that the trader history array is updated properly as well

bweick (approved):

test/protocol/modules/generalIndexModule.spec.ts
cgewecke: The `tradersHistory` is not available on the automatic getter generated by solidity. This is the permissionInfo abi.
```
{
  "inputs": [
    {
      "internalType": "contract ISetToken",
      "name": "",
      "type": "address"
    }
  ],
  "name": "permissionInfo",
  "outputs": [
    {
      "internalType": "bool",
      "name": "anyoneTrade",
      "type": "bool"
    }
  ],
  "stateMutability": "view",
  "type": "function",
  "gas": "0x81b320"
},
```

cgewecke: Only thing we can read out of that without writing a getter is `anyoneTrade`. Should we write one?

contracts/protocol/modules/GeneralIndexModule.sol
cgewecke: d1ebe6e

test/protocol/modules/generalIndexModule.spec.ts
cgewecke: 770b584

PR #82, opened 2021-04-28
Add initial optimism deps and contracts

cgewecke:
**DO NOT MERGE: PR TARGETS #81** 

PR:
+ selects the initial set of core contracts and test contract deps for Optimism L2
+ adds optimism's hardhat plugin and hardhat sizer dependencies (b/c size limits are critical)
+ add two OpenZeppelin contracts modified to be ovm-compatible:
  + SafeERC20.sol
  + Address.sol
+ adds misc. scripts to get everything running.
+ minimizes the circleci config to build. We will need to add custom setup as it becomes available. Some things like coverage are not currently possible with an OVM client. 

**Additional notes**

The new Zeppelin contracts are ported from [derekbar90/optimism-openzeppelin-compat][1]. Optimism does not support the `BALANCE` opcode. Instead balances are checked by querying an address's native WETH balance on the Optimism network. These contracts are updated to reflect that. 

Optimism's compiler plugin has a bug I've written a patch for here and reported at [eth-optimism/optimism  64][2].

[1]: https://github.com/derekbar90/optimism-openzeppelin-compat/blob/66ef793bc49c948dc9430c289f1f9447a3632fa9/patches/@openzeppelin+contracts+3.4.1.patch
[2]: https://github.com/ethereum-optimism/optimism/issues/674


hardhat.config.ts
```diff
@@ -7,18 +7,26 @@ import "@nomiclabs/hardhat-waffle";
 import "hardhat-typechain";
 import "solidity-coverage";
 import "hardhat-deploy";
+import "hardhat-contract-sizer";
+import "@eth-optimism/plugins/hardhat/compiler";
```
gakonst, line 11: The latest one is `@eth-optimism/hardhat-ovm`: https://www.npmjs.com/package/@eth-optimism/hardhat-ovm

hardhat.config.ts
```diff
@@ -45,19 +53,29 @@ const config: HardhatUserConfig = {
...
     target: "ethers-v5",
   },
   mocha: {
-    timeout: 100000,
+    grep: "@ovm",
+  },
+  paths: {
+    artifacts: OVM ? "./artifacts-ovm" : "./artifacts",
+    cache: OVM ? "./cache-ovm" : "./cache",
   },
```
gakonst, line 77: Should not be needed if you use hardhat-ovm

cgewecke: This project consumes its bytecode out of `typechain` factories rather than hardhat artifacts. The typechain plugin reads `paths.artifacts` [here][1].

Note to self: check what this value is in the typechain plugin compile hook.
 
`cache` definitely removable. 

[1]: https://github.com/ethereum-ts/TypeChain/blob/541b98c6967d9324552a328efe354e8cde41978d/packages/hardhat/src/index.ts#L47

PR #83, opened 2021-04-29
Final review fixes for GeneralIndexModule

cgewecke:
**Review Fixes Summary** 

+ **Note**: Even though external positions are not allowed during rebalance instantiation, external positions can be added during the rebalance itself. 
  + cbcd4d9: Check added to the trade parameter validation method
+ **Line 198**: If we expect this to be reused, consider moving this require into the Position.sol library
  + cbcd4d9: uses Position's `hasExternalPosition` helper in the `require`
+ **Line 245**: nit - add linebreak
  + 95c6809
+ **nit**: Consider moving the lowest level functions  to the bottom of the file 
  + 95c6809: internal functions re-ordered

+ **[Line 204][3]**: Instead of emitting an event per component, what would the advantage / disadvantage be of emitting a  a single event?

> [Chris]:  Uncertain about this...
> [Brian/Felix] (from meeting), yes this would be better as a single event...
 + 63f9b05: Fixed

+  **[Line 859][4]**: Instead of “isSendTokenFixed”, I grasp “isSellComponent / isBuyComponent” more intuitively
_getNormalizedTargetUnit

> [Chris]: There was a lot of back and forth about this in #75 / commit 3f0112d. [Brian argues in this PR comment][1] that buy/sell is more intuitive but fundamentally mis-labels these variables and it's necessary to accept the fixed/floating naming convention (unless there's a 3rd way to clarity).  
  + 3b8d415: Added a clarifying comment / table at the top of the internal methods block diagramming how fixed / floating maps onto the permutations of buying and selling at different stages in the rebalance process: 

**Wont Fix**
+ **[Line 630][2]**: If we can find a way to check the component quicker than O(n), that would be nice. 

> [Chris]: It seems like we'd need a mapping for the rebalanceComponents which has to be kept in sync with the array. To get an upper bound for the impact of optimizing here I ran cost analysis for completely removing the check. Diff was small relative to the overall cost of the trades (with 4 components).

Contract | with check | without check | $ diff |
|----|----|----|----|
|trade|364066|354202| - $2 |
|tradeRemainingWeth|286393|284885| < $1 |

+  **[Line 670][5]**: (Brian) It would be better to revert on an `ethQuantityLimit` which is too large than quietly trade the remainder. Would prefer passing `_ethQuantityLimit` directly here.  
  + Decided not to do this because it's incompatible with Balancer's param validation. 

[1]: https://github.com/SetProtocol/set-protocol-v2/pull/75#issuecomment-824209756
[2]: https://github.com/SetProtocol/set-protocol-v2/blob/eb8801d6cdcad8dbce133d33cd28e774f69b790c/contracts/protocol/modules/GeneralIndexModule.sol#L630
[3]: https://github.com/SetProtocol/set-protocol-v2/blob/eb8801d6cdcad8dbce133d33cd28e774f69b790c/contracts/protocol/modules/GeneralIndexModule.sol#L204
[4]: https://github.com/SetProtocol/set-protocol-v2/blob/eb8801d6cdcad8dbce133d33cd28e774f69b790c/contracts/protocol/modules/GeneralIndexModule.sol#L859
[5]: https://github.com/SetProtocol/set-protocol-v2/blob/eb8801d6cdcad8dbce133d33cd28e774f69b790c/contracts/protocol/modules/GeneralIndexModule.sol#L670

### DIFF
--- a/contracts/protocol/modules/GeneralIndexModule.sol
+++ b/contracts/protocol/modules/GeneralIndexModule.sol
@@ -195,11 +195,7 @@ contract GeneralIndexModule is ModuleBase, ReentrancyGuard {
         );
 
         for (uint256 i = 0; i < aggregateComponents.length; i++) {
-            require(
-                _setToken.getExternalPositionModules(aggregateComponents[i]).length == 0,
-                "External positions not allowed"
-            );
-
+            require(!_setToken.hasExternalPosition(aggregateComponents[i]), "External positions not allowed");
             executionInfo[_setToken][IERC20(aggregateComponents[i])].targetUnit = aggregateTargetUnits[i];
             emit TargetUnitsUpdated(_setToken, aggregateComponents[i], aggregateTargetUnits[i], _positionMultiplier);
         }
@@ -636,6 +632,8 @@ contract GeneralIndexModule is ModuleBase, ReentrancyGuard {
             componentInfo.lastTradeTimestamp.add(componentInfo.coolOffPeriod) <= block.timestamp,
             "Component cool off in progress"
         );
+
+        require(!_setToken.hasExternalPosition(address(_component)), "External positions not allowed");
     }
 
     /**
--- a/contracts/protocol/modules/GeneralIndexModule.sol
+++ b/contracts/protocol/modules/GeneralIndexModule.sol
@@ -106,7 +106,6 @@ contract GeneralIndexModule is ModuleBase, ReentrancyGuard {
 
     /* ============ Events ============ */
 
-    event TargetUnitsUpdated(ISetToken indexed _setToken, address indexed _component, uint256 _newUnit, uint256 _positionMultiplier);
     event TradeMaximumUpdated(ISetToken indexed _setToken, address indexed _component, uint256 _newMaximum);
     event AssetExchangeUpdated(ISetToken indexed _setToken, address indexed _component, string _newExchangeName);
     event CoolOffPeriodUpdated(ISetToken indexed _setToken, address indexed _component, uint256 _newCoolOffPeriod);
@@ -128,7 +127,12 @@ contract GeneralIndexModule is ModuleBase, ReentrancyGuard {
         uint256 _protocolFee
     );
 
-    event RebalanceStarted(ISetToken indexed _setToken);
+    event RebalanceStarted(
+        ISetToken indexed _setToken,
+        address[] aggregateComponents,
+        uint256[] aggregateTargetUnits,
+        uint256 indexed positionMultiplier
+    );
 
     /* ============ Constants ============ */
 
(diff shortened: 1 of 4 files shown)


contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -617,6 +617,51 @@ contract GeneralIndexModule is ModuleBase, ReentrancyGuard {
...
+     *        Step 1        |       Step 2
+     * -------------------------------------------
+     * Component --> WETH   |   WETH --> Component
+     * -------------------------------------------
+     *
+     * The syntax we use frames this as trading from a "fixed" amount of one component to a
+     * "fixed" amount of another via a "floating limit" which is *either* the maximum size of
+     * the trade we want to make (trades may be tranched to avoid moving markets) OR the minimum
+     * amount of tokens we expect to receive. The different meanings of the floating limit map to
+     * the trade sequence as below:
```
cgewecke, line 634: Maybe: "the minimum amount of token we expect to receive to acheive our target."

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -617,6 +617,51 @@ contract GeneralIndexModule is ModuleBase, ReentrancyGuard {
...
+     *
+     * Step 1: Component --> WETH
+     * ----------------------------------------------------------
+     *                     | Fixed |     Floating limit         |
+     * ----------------------------------------------------------
+     * send  (Component)   |  Yes  |                            |
+     * recieve (WETH)      |       |   Maximum trade size       |
```
bweick, line 641: This should be minimum receive amount

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -617,6 +617,51 @@ contract GeneralIndexModule is ModuleBase, ReentrancyGuard {
...
+     * ----------------------------------------------------------
+     *
+     * Step 2: WETH --> Component
+     * ----------------------------------------------------------
+     *                     | Fixed  |    Floating limit         |
+     * ----------------------------------------------------------
+     * send  (WETH)        |   NO   |                           |
+     * recieve (Component) |        |  Min amount to receive    |
```
bweick, line 649: Would note that the component is fixed here, and the floating limit is the maximum amount of ETH you are willing to send to the exchange

contracts/protocol/modules/GeneralIndexModule.sol
```diff
@@ -617,6 +617,51 @@ contract GeneralIndexModule is ModuleBase, ReentrancyGuard {
...
+     * ----------------------------------------------------------
+     *
+     * Additionally, there is an edge case where price volatility during a rebalance
+     * results in remaindered WETH which needs to be allocated proportionately. In this case
```
bweick, line 653: remaindered --> remaining

contracts/protocol/modules/GeneralIndexModule.sol
cgewecke: Oh this is backwards!

bweick (approved):

bweick:
Let's bump the package number here as well

PR #84, opened 2021-04-29
Restructure SetToken to reduce size & make deployable to ovm

cgewecke:
**DO NOT MERGE: TARGETS #82**

(Very WIP) 
+ Moves SetToken public getters and some internal logic into externally linked libraries.
+ Makes all reason strings < 32 bytes

This gets us just under the effective contract size limit (as measured in practice when trying to deploy to Optimism client)

Reduces ovm compiled size of SetTokenCreator from 24.1KB to 22.16KB.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,9 +1,12 @@
 version: 2
 
 jobs:
-  install:
+  checkout_and_compile:
     docker:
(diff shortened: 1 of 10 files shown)


cgewecke:
Closing in favor of #85

PR #85, opened 2021-04-30
Add optimism client to CI and test set core deployment

cgewecke:
**DO NOT MERGE: TARGETS #84**

+ Adds Optimism client in CI
+ Adds tests to validate core system deploys successfully to L2


external/contracts/openzeppelin/utils/Address.sol
```diff
@@ -133,12 +147,16 @@ library Address {
     function functionCallWithValue(address target, bytes memory data, uint256 value, string memory errorMessage) internal returns (bytes memory) {
         // >>>> OPTIMISM:START_CHANGE
         require(WETHBalance(address(this)) >= value, "Address: insufficient balance for call");
-        // <<<< OPTIMISM:END_CHANGE
         require(isContract(target), "Address: call to non-contract");
 
+        if (value > 0){
+            bool success = WETHTransfer(target, value);
+            _verifyCallResult(success, "", "Address: WETH transfer failed");
+        }
```
cgewecke, line 155: Should go after call.

cgewecke: Actually, need to investigate how `value` gets compiled down...current order might be right. 

Also double check current Optimism's client's management of this to see if necessary.

cgewecke:
Closing for the moment ... will revisit when support for Optimism is scheduled.

PR #87, opened 2021-05-06
Add AXS migration wrap adapter

Alpha-Guy:
Old AXS Token: [0xF5D669627376EBd411E34b98F19C868c8ABA5ADA](https://etherscan.io/address/0xf5d669627376ebd411e34b98f19c868c8aba5ada)
New AXS Token: [0xBB0E17EF65F82Ab018d8EDd776e8DD940327B28b](https://etherscan.io/address/0xbb0e17ef65f82ab018d8edd776e8dd940327b28b)
Contract facilitating the migration: [TokenSwap contract](https://etherscan.io/address/0x73b1714fb3bfaefa12f3707befcba3205f9a1162#code)

Function used for migration: `swapToken()`
Note: Above function does not take in amount of tokens to be swapped and by default swaps the entire old AXS balance of the caller for the new AXS tokens.


bweick:
Realizing there may be an issue here. So we will be approving an amount of tokens equal to `totalSupply.preciseMul(positionUnit)` however this migrator tries to `transferFrom` the entire balance which probably will result in an error because there's always some extra wei sitting around due to rounding errors. Let's think about ways we can work around that.

contracts/protocol/integration/wrap/AxieInfinityMigrationWrapAdapter.sol
```diff
@@ -0,0 +1,139 @@
...
+
+interface ITokenSwap {
+    function swapToken() external;    
+}
+
+/**
+ * @title AxieInfinityMigrationWrapAdapter
+ * @author Set Protocol
+ *
+ * Wrap adapter for one time token migration that returns data for wrapping old AXS token into new AXS token.
```
bweick, line 32: Let's make a note that this adapter will be involved in more than just creating data. Perhaps even a list of all the interactions between the module, adapter, and token swapping contract (I'd include where approvals are going to as well).

contracts/protocol/integration/wrap/AxieInfinityMigrationWrapAdapter.sol
```diff
@@ -0,0 +1,139 @@
...
+        uint256 _underlyingUnits
+    )
+        external
+        view
+        returns (address, uint256, bytes memory)
+    {
+        require(_underlyingToken == oldToken, "Must be old AXS token");
+        require(_wrappedToken == newToken, "Must be new AXS token");
+
+        // swapToken()
```
bweick, line 110: `swapToken(uint256)` also let's note that this interface is the interface on this adapter and not the one on the `tokenSwap` contract

contracts/protocol/integration/wrap/AxieInfinityMigrationWrapAdapter.sol
```diff
@@ -0,0 +1,139 @@
...
+    /* ============ External Functions ============ */
+
+    /**
+     * Pulls specified amount of old AXS tokens from the `msg.sender` and swaps them for new AXS tokens 
+     * for a 1:1 ratio via the Axie TokenSwap contract. Transfers the received amount of new AXS tokens
+     * back to the `msg.sender`
+     *
+     * @param _amount           Total amount of old AXS tokens to be swapped for new AXS tokens
+     */
+    function swapToken(uint256 _amount) external {
```
bweick, line 78: Do we want to name this something a bit more descriptive instead of the same name used on the AXS contract? Maybe something like `swapTokensUsingAdapter`

contracts/protocol/integration/wrap/AxieInfinityMigrationWrapAdapter.sol
```diff
@@ -0,0 +1,173 @@
...
+
+interface ITokenSwap {
+    function swapToken() external;    
+}
+
+/**
+ * @title AxieInfinityMigrationWrapAdapter
+ * @author Set Protocol
+ *
+ * Wrap adapter for one time migration to new AXS tokens. This adapter contract acts as a middleman in the token migration process.
```
bweick, line 32: This is awesome 💯 . Well done!

bweick (approved):

bweick:
Let's just bump the package here then we're good to go!

PR #88, opened 2021-05-08
Adding CompClaimAdapter

bronco-eth:
This is a simple claim adapter that allows managers to claim COMP from assets deposited on Compound.

It's interfacing with Comptroller and conforms to the IClaimAdapter interface used by the Set Protocol V2 ClaimModule.

Tests are using mocked Comptroller contract and are limited to ClaimAdapter <> ClaimModule integration.

### DIFF
--- a/test/integration/claim/compClaimAdapter.spec.ts
+++ b/test/integration/claim/compClaimAdapter.spec.ts
@@ -1,0 +1,183 @@
+import "module-alias/register";
+import { waffle } from "hardhat";
+import { Contract, BigNumber } from "ethers";
+import { Address } from "@utils/types";
+import { Account } from "@utils/test/types";
+import { ADDRESS_ZERO } from "@utils/constants";
+import {
+  CompClaimAdapter,
+  ClaimModule,
+  SetToken,
+  StandardTokenMock,
+} from "@utils/contracts";
+import DeployHelper from "@utils/deploys";
+import { ether } from "@utils/index";
+import {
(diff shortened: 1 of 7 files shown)


test/integration/claim/compClaimAdapter.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+      function subject(): Promise<any> {
+        return claimModule.connect(owner.wallet).claim(setToken.address, comptroller.address, compClaimAdapterIntegrationName);
+      }
+
+      it("should claim accrued reward", async function() {
+        const claim = await subject();
+        const receipt = await claim.wait();
+
+        // Get RewardClaimed event dispatched in a ClaimModule#_claim call
+        const rewardClaimed: any = receipt.events.find((e: any): any => e.event == "RewardClaimed");
```
richardliang, line 159: So the actual claimComp doesn't transfer any tokens? Ideally it would be a mock to transfer reward token to the SetToken and check the balance of:
https://github.com/SetProtocol/set-protocol-v2/blob/master/test/protocol/modules/claimModule.spec.ts#L970

bronco-eth: Good point, I used ComptrollerMock that was already in the repo. It's claiming function is triggering a transfer on a mocked Comp ERC20 token. New test is added below and it's checking how the balance changes after the #claim is executed.

richardliang (approved):
LGTM

PR #92, opened 2021-05-27
Update package for public npm (make ethers a non-dev dep)

cgewecke:
This makes sure `ethers` packages are included in the published package (they're required for dist/typechain and some utils) by moving them to regular dependencies.

Have pinned these deps since the exact versions we have yarn.locked [are required for everything to work together correctly][1]. A user who installed with floating versions would get whatever latest is within the range and run into tsc errors.

Also added: 
+ a peer dependencies section to `package.json` for hardhat / waffle deps required by the utils/fixtures
+ README section about npm distro
+ changes to lifecycle scripts so ABIs don't ship with hardcoded gas estimates

[1]: https://app.circleci.com/pipelines/github/SetProtocol/set-protocol-v2/492/workflows/3d4204b1-8d89-45cd-9040-a85cb4442998/jobs/1684


asoong (approved):

asoong (approved):
LGTM

PR #95, opened 2021-06-03
Add Uniswap V3 Fixture

ncitron:


utils/fixtures/uniswapV3Fixture.ts
```diff
@@ -0,0 +1,99 @@
...
+  ): Promise<void> {
+
+    if (_token0.toLowerCase() > _token1.toLowerCase()) {
+      [ _amount0, _amount1 ] = [ _amount1, _amount0 ];
+    }
+
+    [ _token0, _token1 ] = this.getTokenOrder(_token0, _token1);
+
+    const tickSpacing = _fee / 50;
+    const maxTick = Math.floor(887272 / tickSpacing) * tickSpacing;
```
richardliang, line 73: What's this number? `887272`

richardliang: Should leave a comment describing these numbers

utils/fixtures/uniswapV3Fixture.ts
```diff
@@ -0,0 +1,99 @@
...
+    _recipient: Address
+  ): Promise<void> {
+
+    if (_token0.toLowerCase() > _token1.toLowerCase()) {
+      [ _amount0, _amount1 ] = [ _amount1, _amount0 ];
+    }
+
+    [ _token0, _token1 ] = this.getTokenOrder(_token0, _token1);
+
+    const tickSpacing = _fee / 50;
```
richardliang, line 72: Comment describing why fee is divided by 50

utils/fixtures/uniswapV3Fixture.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  public async createNewPair(_token0: Address, _token1: Address, _fee: BigNumberish, _sqrtPriceX96: BigNumberish): Promise<UniswapV3Pool> {
+    [ _token0, _token1 ] = this.getTokenOrder(_token0, _token1);
+
+    await this.nftPositionManager.createAndInitializePoolIfNecessary(_token0, _token1, _fee, _sqrtPriceX96);
+
+    return this.getPool(_token0, _token1, _fee);
+  }
+
+  public async addLiquidityWide(
```
richardliang, line 57: How wide is this? From 0 to infinity?

utils/fixtures/uniswapV3Fixture.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    [ _token0, _token1 ] = this.getTokenOrder(_token0, _token1);
+
+    const tickSpacing = _fee / 50;
+    const maxTick = Math.floor(887272 / tickSpacing) * tickSpacing;
+    const minTick = Math.ceil(-maxTick / tickSpacing) * tickSpacing;
+
+    await this.nftPositionManager.connect(this._ownerSigner).mint({
+      fee: _fee,
+      token0: _token0,
```
richardliang, line 78: So order matters when minting?

test/fixtures/uniswapV3.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+      await subjectTokenOne.approve(uniswapV3Fixture.nftPositionManager.address, subjectAmountOne);
+      await subjectTokenTwo.approve(uniswapV3Fixture.nftPositionManager.address, subjectAmountTwo);
+    });
+
+    it("should add liquidity into the pool", async () => {
+      await subject();
+
+      const pool = await uniswapV3Fixture.getPool(subjectTokenOne.address, subjectTokenTwo.address, subjectFee);
+
+      expect(await subjectTokenOne.balanceOf(pool.address)).to.be.gt(0);
```
richardliang, line 136: Can we calculate the exact balance here?

richardliang (commented):
Mostly looks good, just need some more comments on how some of the liquidity calculations work

utils/fixtures/uniswapV3Fixture.ts
ncitron: Yep. Should I maybe add a javadoc comment for this method?

utils/fixtures/uniswapV3Fixture.ts
ncitron: Yep

utils/fixtures/uniswapV3Fixture.ts
```diff
@@ -0,0 +1,150 @@
...
+   * @param _dai    dai address
+   */
+  public async initialize(_owner: Account, _weth: Address, _wbtc: Address, _dai: Address): Promise<void> {
+    this.factory = await this._deployer.external.deployUniswapV3Factory();
+    this.swapRouter = await this._deployer.external.deploySwapRouter(this.factory.address, _weth);
+    this.nftDescriptor = await this._deployer.external.deployNFTDescriptor();
+    this.nftPositionManager = await this._deployer.external.deployNftPositionManager(this.factory.address, _weth, this.nftDescriptor.address);
+    this.quoter = await this._deployer.external.deployQuoter(this.factory.address, _weth);
+
+    this.wethDaiPool = await this.createNewPair(_weth, _dai, 3000, BigNumber.from("1490848768284521510823413542"));
```
bweick, line 59: Ideally here we pass in initial prices and calculate the `sqrtPriceX96` from those. Also if calculating `sqrtPriceX96` is a common thing should just make a helper for that. For now can be an internal function in this class

utils/fixtures/uniswapV3Fixture.ts
```diff
@@ -0,0 +1,150 @@
...
+  /**
+   * Creates and initializes a new pool
+   *
+   * @param _token0         address of the first token
+   * @param _token1         address of the second token
+   * @param _fee            fee tier of either 500, 3000, or 10000
+   * @param _sqrtPriceX96   the initial price parameter equal to sqrt(priceToken1 / priceToken0) * 2^96
+   * @returns               a new Uniswap V3 pool
+   */
+  public async createNewPair(_token0: Address, _token1: Address, _fee: BigNumberish, _sqrtPriceX96: BigNumberish): Promise<UniswapV3Pool> {
```
bweick, line 72: Noting here re: comment above we prob want to change this functions interface

utils/fixtures/uniswapV3Fixture.ts
```diff
@@ -0,0 +1,150 @@
...
+  public async addLiquidityWide(
+    _token0: Address,
+    _token1: Address,
+    _fee: number,
+    _amount0: BigNumber,
+    _amount1: BigNumber,
+    _recipient: Address
+  ): Promise<void> {
+
+    if (_token0.toLowerCase() > _token1.toLowerCase()) {
```
bweick, line 99: There's some confusing naming going on here. Do we need to redeclare some variables? Seems like we can pass in a _token0 and have that set to _token1?

bweick: Let's declare some `token0Ordered` and `token1Ordered` vars separate from `_token0` and `_token1` so we aren't overwriting params we pass in. Same for `amounts`, though not sure the descriptor we want to use for that one

test/fixtures/uniswapV3.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  describe("#addLiquidityWide", async () => {
+
+    let subjectTokenOne: StandardTokenMock;
+    let subjectTokenTwo: StandardTokenMock | WETH9;
+    let subjectFee: number;
+    let subjectAmountOne: BigNumber;
+    let subjectAmountTwo: BigNumber;
+
+    async function subject(): Promise<void> {
```
bweick, line 107: Typically we place the `subject` function after the `beforeEach`

test/fixtures/uniswapV3.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+  });
+
+  describe("#createPool", async () => {
+
+    let subjectTokenOne: StandardTokenMock;
+    let subjectTokenTwo: StandardTokenMock;
+    let subjectFee: BigNumberish;
+    let subjectSqrtPriceX96: BigNumber;
+
+    async function subject(): Promise<UniswapV3Pool> {
```
bweick, line 78: `subject` after `beforeEach`

bweick (commented):
Good stuff, really close, one thing to keep in mind with fixtures is that you'll be using them to write tests so you want to make them as painless to use as possible. So really think about usability, as an example having to calculate or come up with a `sqrtPriceX96` every time would be really annoying. Even reading the tests it's kind've unclear what price is being used for some of them.

utils/fixtures/uniswapV3Fixture.ts
```diff
@@ -49,15 +55,15 @@ export class UniswapV3Fixture {
    * @param _wbtc   wbtc address
    * @param _dai    dai address
    */
-  public async initialize(_owner: Account, _weth: Address, _wbtc: Address, _dai: Address): Promise<void> {
+  public async initialize(_owner: Account, _weth: Token, _wbtc: Token, _dai: Token): Promise<void> {
```
bweick, line 58: Wonder if it makes sense to just have the prices passed in here. Oftentimes there are oracles that are also a part of a system set up so would be good to match the oracles and the pools explicitly

bweick (approved):

PR #97, opened 2021-06-10
ClaimModule Audit fixes

ncitron:

### DIFF
--- a/contracts/protocol/modules/ClaimModule.sol
+++ b/contracts/protocol/modules/ClaimModule.sol
@@ -262,6 +262,7 @@ contract ClaimModule is ModuleBase {
     function removeModule() external override {
         delete anyoneClaim[ISetToken(msg.sender)];
 
+        // explicitly delete all elements for gas refund
         address[] memory setTokenPoolList = rewardPoolList[ISetToken(msg.sender)];
         for (uint256 i = 0; i < setTokenPoolList.length; i++) {
 
@@ -430,7 +431,7 @@ contract ClaimModule is ModuleBase {
         _rewardPoolClaimSettings.push(adapter);
         claimSettingsStatus[_setToken][_rewardPool][adapter] = true;
 
-        if (_rewardPoolClaimSettings.length == 1) {
+        if (!rewardPoolStatus[_setToken][_rewardPool]) {
             rewardPoolList[_setToken].push(_rewardPool);
             rewardPoolStatus[_setToken][_rewardPool] = true;
         }
@@ -479,11 +480,6 @@ contract ClaimModule is ModuleBase {
         if (claimSettings[_setToken][_rewardPool].length == 0) {
             rewardPoolList[_setToken] = rewardPoolList[_setToken].remove(_rewardPool);
             rewardPoolStatus[_setToken][_rewardPool] = false;
-
-            for (uint256 i = 0; i < claimSettings[_setToken][_rewardPool].length; i++) {
-                delete claimSettings[_setToken][_rewardPool][i];
-            }
-            delete claimSettings[_setToken][_rewardPool];
         }
     }
 
--- a/contracts/protocol/modules/ClaimModule.sol
+++ b/contracts/protocol/modules/ClaimModule.sol
@@ -277,6 +277,9 @@ contract ClaimModule is ModuleBase {
         }
         
         for (uint256 i = 0; i < rewardPoolList[ISetToken(msg.sender)].length; i++) {
+            address toRemove = rewardPoolList[ISetToken(msg.sender)][i];
+            rewardPoolStatus[ISetToken(msg.sender)][toRemove] = false;
+            
             delete rewardPoolList[ISetToken(msg.sender)][i];
         }
         delete rewardPoolList[ISetToken(msg.sender)];
--- a/contracts/protocol/modules/ClaimModule.sol
+++ b/contracts/protocol/modules/ClaimModule.sol
@@ -471,14 +471,13 @@ contract ClaimModule is ModuleBase {
      */
     function _removeClaim(ISetToken _setToken, address _rewardPool, string calldata _integrationName) internal {
         address adapter = getAndValidateAdapter(_integrationName);
-        address[] memory _rewardPoolClaimSettings = claimSettings[_setToken][_rewardPool];
 
         require(claimSettingsStatus[_setToken][_rewardPool][adapter], "Integration must be added");
-        claimSettings[_setToken][_rewardPool] = _rewardPoolClaimSettings.remove(adapter);
+        claimSettings[_setToken][_rewardPool].removeStorage(adapter);
         claimSettingsStatus[_setToken][_rewardPool][adapter] = false;
 
         if (claimSettings[_setToken][_rewardPool].length == 0) {
-            rewardPoolList[_setToken] = rewardPoolList[_setToken].remove(_rewardPool);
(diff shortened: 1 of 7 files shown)


contracts/protocol/modules/ClaimModule.sol
```diff
@@ -439,11 +469,17 @@ contract ClaimModule is ModuleBase {
...
-        require(_rewardPoolClaimSettings.contains(adapter), "Integration must be added");
+        require(claimSettingsStatus[_setToken][_rewardPool][adapter], "Integration must be added");
         claimSettings[_setToken][_rewardPool] = _rewardPoolClaimSettings.remove(adapter);
+        claimSettingsStatus[_setToken][_rewardPool][adapter] = false;
 
         if (claimSettings[_setToken][_rewardPool].length == 0) {
             rewardPoolList[_setToken] = rewardPoolList[_setToken].remove(_rewardPool);
+            rewardPoolStatus[_setToken][_rewardPool] = false;
+
+            for (uint256 i = 0; i < claimSettings[_setToken][_rewardPool].length; i++) {
```
bweick, line 480: Just leave a comment we're deleting all array items in order to get the gas back

contracts/protocol/modules/ClaimModule.sol
```diff
@@ -254,8 +264,21 @@ contract ClaimModule is ModuleBase {
 
         address[] memory setTokenPoolList = rewardPoolList[ISetToken(msg.sender)];
         for (uint256 i = 0; i < setTokenPoolList.length; i++) {
+
```
bweick, line 267: Add some note that we are deleting all the items in the arrays for the gas refund

bweick (commented):
For the tests we also want to make sure that all the mappings have been updated as we expect too. Right now there's no checks to make sure that the mappings that were added were properly updated.

contracts/protocol/modules/ClaimModule.sol
bweick: Actually we don't need this right? Don't we already know this is equal to length 0 based on the if statement it's included in?

contracts/protocol/modules/ClaimModule.sol
```diff
@@ -439,11 +469,17 @@ contract ClaimModule is ModuleBase {
         address adapter = getAndValidateAdapter(_integrationName);
         address[] memory _rewardPoolClaimSettings = claimSettings[_setToken][_rewardPool];
 
-        require(_rewardPoolClaimSettings.contains(adapter), "Integration must be added");
+        require(claimSettingsStatus[_setToken][_rewardPool][adapter], "Integration must be added");
         claimSettings[_setToken][_rewardPool] = _rewardPoolClaimSettings.remove(adapter);
```
bweick, line 473: I think we may want to use `removeStorage` here. May have been missed but this is inefficient

contracts/protocol/modules/ClaimModule.sol
```diff
@@ -439,11 +469,17 @@ contract ClaimModule is ModuleBase {
...
+        claimSettingsStatus[_setToken][_rewardPool][adapter] = false;
 
         if (claimSettings[_setToken][_rewardPool].length == 0) {
             rewardPoolList[_setToken] = rewardPoolList[_setToken].remove(_rewardPool);
```
bweick, line 477: Use `removeStorage` here

contracts/protocol/modules/ClaimModule.sol
```diff
@@ -439,11 +469,17 @@ contract ClaimModule is ModuleBase {
...
+            rewardPoolStatus[_setToken][_rewardPool] = false;
+
+            for (uint256 i = 0; i < claimSettings[_setToken][_rewardPool].length; i++) {
+                delete claimSettings[_setToken][_rewardPool][i];
+            }
             delete claimSettings[_setToken][_rewardPool];
```
bweick, line 483: Do we gain anything by deleting this given that we know it has 0 length? Genuinely don't know, maybe frees up some storage space?

contracts/protocol/modules/ClaimModule.sol
```diff
@@ -395,11 +423,13 @@ contract ClaimModule is ModuleBase {
         address adapter = getAndValidateAdapter(_integrationName);
         address[] storage _rewardPoolClaimSettings = claimSettings[_setToken][_rewardPool];
 
-        require(!_rewardPoolClaimSettings.contains(adapter), "Integration names must be unique");
+        require(!claimSettingsStatus[_setToken][_rewardPool][adapter], "Integration names must be unique");
         _rewardPoolClaimSettings.push(adapter);
+        claimSettingsStatus[_setToken][_rewardPool][adapter] = true;
 
         if (_rewardPoolClaimSettings.length == 1) {
```
bweick, line 430: I think we can actually make this cleaner just by checking to see if `rewardPoolStatus[_setToken][_rewardPool] = true`. Before we were using the arrays as the source of truth since we didn't have the mappings. But now with the mappings we can make explicit checks and probably simplify the logic. 

Same can probably be done for `_removeClaim`

contracts/protocol/modules/ClaimModule.sol
```diff
@@ -71,11 +78,14 @@ contract ClaimModule is ModuleBase {
...
 
-    // Array of rewardPool addresses to claim rewards for the SetToken
+    // Map and array of rewardPool addresses to claim rewards for the SetToken
     mapping(ISetToken => address[]) public rewardPoolList;
+    mapping(ISetToken => mapping(address => bool)) public rewardPoolStatus;
 
-    // Array of adapters associated to the rewardPool for the SetToken
+    // Map and array of adapters associated to the rewardPool for the SetToken
     mapping(ISetToken => mapping(address => address[])) public claimSettings;
+    mapping(ISetToken => mapping(address => mapping(address => bool))) public claimSettingsStatus;
```
richardliang, line 87: Can you add a comment for what this mapping does and why? Crazy nesting

contracts/protocol/modules/ClaimModule.sol
```diff
@@ -71,11 +78,14 @@ contract ClaimModule is ModuleBase {
     // Indicates if any address can call claim or just the manager of the SetToken
     mapping(ISetToken => bool) public anyoneClaim;
 
-    // Array of rewardPool addresses to claim rewards for the SetToken
+    // Map and array of rewardPool addresses to claim rewards for the SetToken
     mapping(ISetToken => address[]) public rewardPoolList;
+    mapping(ISetToken => mapping(address => bool)) public rewardPoolStatus;
```
richardliang, line 83: Can you add a comment for what this mapping does and why?

contracts/protocol/modules/ClaimModule.sol
```diff
@@ -346,40 +372,46 @@ contract ClaimModule is ModuleBase {
...
         ) = adapter.getClaimCallData(
-            address(_setToken),
+            _setToken,
             _rewardPool
         );
 
         _setToken.invoke(callTarget, callValue, callByteData);
 
+        uint256 finalRewardsBalance = rewardsToken.balanceOf(address(_setToken));
+        uint256 rewards = finalRewardsBalance.sub(initRewardsBalance);
```
richardliang, line 392: we can directly pass `finalRewardsBalance.sub(initRewardsBalance)` into the event below and eliminate this variable

richardliang (approved):

richardliang (approved):
Reviewed against the auditor report and @bweick's previous comments and LGTM. Left some notes around minor comments

PR #99, opened 2021-06-17
Add TradeSplitter

ncitron:
# Code Review Processes
## New Feature Review
* Design doc is created and posted here: https://github.com/SetProtocol/STIPS/blob/ncitron/trade-splitter/STIPS/STIP-002.md

README Checks

Code Checks
- [] Add javadocs. 

Broader Considerations


contracts/product/TradeSplitter.sol
```diff
@@ -0,0 +1,272 @@
...
+
+            address sushiPairA = sushiFactory.getPair(_path[0], _path[1]);
+            address sushiPairB = sushiFactory.getPair(_path[1], _path[2]);
+
+            uint256 sushiValueA = ERC20(_path[1]).balanceOf(sushiPairA);
+            uint256 sushiValueB = ERC20(_path[1]).balanceOf(sushiPairB);
+
+            if(sushiValueA == 0 || sushiValueB == 0) return (_size, 0);
+
+            uint256 ratio = sushiValueA.add(sushiValueB).preciseMul(uniValueA).preciseMul(uniValueB)
```
richardliang, line 216: Let's break this down for better readability

contracts/product/TradeSplitter.sol
```diff
@@ -0,0 +1,272 @@
...
+    )
+        external
+        returns (uint256 totalOutput)
+    {
+        require(_path.length <= 3 && _path.length != 0, "TradeSplitter: incorrect path length");
+
+        ERC20 inputToken = ERC20(_path[0]);
+        inputToken.transferFrom(msg.sender, address(this), _amountIn);
+        
+        (uint256 uniTradeSize, uint256 sushiTradeSize) = _getTradeSizes(_path, _amountIn);
```
richardliang, line 94: It may be cleaner to refactor into a unified getter struct for every variable we need since these variables are reused across every public function. We can then pass the struct around. Similar to how we do it in the TradeModule and FLIStrategy contracts

https://github.com/SetProtocol/index-coop-smart-contracts/blob/master/contracts/adapters/FlexibleLeverageStrategyAdapter.sol#L231

This may include getting allowances in _checkApprovals() too. We also dedupe a lot of the logic in fetching in getTradeSizes

contracts/product/TradeSplitter.sol
```diff
@@ -0,0 +1,272 @@
...
+        uint256 _deadline
+    )
+        external
+        returns (uint256 totalInput)
+    {
+        require(_path.length <= 3 && _path.length != 0, "TradeSplitter: incorrect path length");
+
+        (uint256 uniTradeSize, uint256 sushiTradeSize) = _getTradeSizes(_path, _amountOut);
+
+        uint256 expectedUniInput = _getTradeInputOrOutput(uniRouter, uniTradeSize, _path, false);
```
richardliang, line 130: Shouldn't we be transferring amountInMax? And refunding the tokens to the SetToken that are unused at the end of the swap.

contracts/product/TradeSplitter.sol
```diff
@@ -0,0 +1,272 @@
...
+        uint256 expectedSushiInput = _getTradeInputOrOutput(sushiRouter, sushiTradeSize, _path, false);
```
richardliang, line 131: Since `expectedUniInput` is always equal to actual `uniInput` and `expectedUniInput` to `sushiInput` we can require `uniInput` + `sushiInput` < `_amountInMax` and revert earlier before transferring tokens in

contracts/product/AMMSplitter.sol
```diff
@@ -0,0 +1,367 @@
...
+        TradeInfo memory tradeInfo = _getTradeSizes(_path, _amountOut);
+
+        uint256 expectedUniInput = _getTradeInputOrOutput(uniRouter, tradeInfo.uniSize, _path, false)[0];
+        uint256 expectedSushiInput = _getTradeInputOrOutput(sushiRouter, tradeInfo.sushiSize, _path, false)[0];
+
+        totalInput = expectedUniInput.add(expectedSushiInput);
+        require(totalInput <= _amountInMax, "TradeSplitter: INSUFFICIENT_INPUT_AMOUNT");
+
+        ERC20 inputToken = ERC20(_path[0]);
+        inputToken.transferFrom(msg.sender, address(this), expectedUniInput.add(expectedSushiInput));
```
bweick, line 146: Can use `totalInput` instead of summing again

contracts/product/AMMSplitter.sol
```diff
@@ -0,0 +1,367 @@
...
+
+            address sushiPair = sushiFactory.getPair(_path[0], _path[1]);
+            uint256 sushiValue = ERC20(_path[0]).balanceOf(sushiPair);
+
+            uniPercentage = uniValue.preciseDiv(uniValue.add(sushiValue));
+        } else {
+            address uniPairA = uniFactory.getPair(_path[0], _path[1]);
+            address uniPairB = uniFactory.getPair(_path[1], _path[2]);
+
+            uint256 uniValueA = ERC20(_path[1]).balanceOf(uniPairA);
```
bweick, line 230: Make note that you're taking the balance of the intermediate asset here. Even better if there's an explanation for why

contracts/product/AMMSplitter.sol
```diff
@@ -0,0 +1,367 @@
...
+            uint256 uniValueB = ERC20(_path[1]).balanceOf(uniPairB);
+
+            if(uniValueA == 0 || uniValueB == 0) return TradeInfo(0, _size);
```
bweick, line 233: Should we be concerned if they're both equal to 0 (however unlikely)? I assume that would just revert.

contracts/product/AMMSplitter.sol
```diff
@@ -0,0 +1,367 @@
...
+        uint256 _uniValueA,
+        uint256 _uniValueB,
+        uint256 _sushiValueA,
+        uint256 _sushiValueB
+    ) 
+        internal
+        pure
+        returns (uint256)
+    {
+        uint256 a = _sushiValueA.add(_sushiValueB).preciseMul(_uniValueA).preciseMul(_uniValueB);
```
bweick, line 284: I don't think this will work for low decimal tokens. Imagine the token has only 4 decimals then it will be rounded down to zero after the first `preciseMul`.

I'm not sure that preciseMul is really necessary since you'll be comparing against like units anyway you may be able to just use `.mul`. Let's think about that.

contracts/product/AMMSplitter.sol
```diff
@@ -0,0 +1,367 @@
...
+     * @param _size         the total size of the trade
+     *
+     * @return TradeInfo    TradeInfo struct containing Uniswap and Sushiswap tarde sizes
+     */
+    function _getTradeSizes(address[] calldata _path, uint256 _size) internal view returns (TradeInfo memory) {
+
+        uint256 uniPercentage;
+        if (_path.length == 2) {
+
+            address uniPair = uniFactory.getPair(_path[0], _path[1]);
```
bweick, line 219: Feels like we can clean up a lot of duplicate logic here by having an internal helper function that something like `getPairBalance` (probably can think of a better name) and pass in the token you're looking to get the balanceOf, the other token in the pair, and the router

contracts/product/AMMSplitter.sol
```diff
@@ -0,0 +1,367 @@
...
+            address sushiPairA = sushiFactory.getPair(_path[0], _path[1]);
+            address sushiPairB = sushiFactory.getPair(_path[1], _path[2]);
+
+            uint256 sushiValueA = ERC20(_path[1]).balanceOf(sushiPairA);
+            uint256 sushiValueB = ERC20(_path[1]).balanceOf(sushiPairB);
+
+            if(sushiValueA == 0 || sushiValueB == 0) return TradeInfo(_size, 0);
+
+            uint256 ratio = _calculateTwoHopRatio(uniValueA, uniValueB, sushiValueA, sushiValueB);
+            uniPercentage = ratio.preciseDiv(ratio.add(PreciseUnitMath.PRECISE_UNIT));
```
bweick, line 244: Let's just add a javadoc here briefly deriving going from ratio -> percentage

contracts/product/AMMSplitter.sol
bweick: Maybe `getTokenBalanceInPair`

contracts/product/AMMSplitter.sol
```diff
@@ -0,0 +1,367 @@
...
+        uint256 uniPercentage;
+        if (_path.length == 2) {
+
+            address uniPair = uniFactory.getPair(_path[0], _path[1]);
+            uint256 uniValue = ERC20(_path[0]).balanceOf(uniPair);
+
+            address sushiPair = sushiFactory.getPair(_path[0], _path[1]);
+            uint256 sushiValue = ERC20(_path[0]).balanceOf(sushiPair);
+
+            uniPercentage = uniValue.preciseDiv(uniValue.add(sushiValue));
```
bweick, line 225: What if the token isn't 18 decimals?

contracts/product/AMMSplitter.sol
```diff
@@ -0,0 +1,367 @@
...
+        uint256[] memory sushiTradeResults = _getTradeInputOrOutput(sushiRouter, tradeInfo.sushiSize, _path, _isExactInput);
+
+        amounts = new uint256[](_path.length);
+        for (uint256 i = 0; i < amounts.length; i++) {
+            amounts[i] = uniTradeResults[i].add(sushiTradeResults[i]);
+        }
+    }
+
+    /**
+     * Calculates the optimal trade sizes for Uniswap and Sushiswap.
```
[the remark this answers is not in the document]
ncitron: If the pool isn't balanced it would mess up our routing, but would also cause a revert due to the slippage check. I think the threat here is similar to that of a normal sandwich attack.

contracts/product/AMMSplitter.sol
ncitron: Yeah, it will just end up reverting when it tried to execute the trade, which would be the expected behavior anyway.

contracts/product/AMMSplitter.sol
ncitron: This is safe since both tokens are going to be measured in the same units. We actually need to use preciseDiv here so the outputs are normalized to 1e18.

contracts/product/AMMSplitter.sol
```diff
@@ -294,15 +296,29 @@ contract AMMSplitter {
      * @param _sushiAmount  Sushiswap input amount
      * @param _token        Token being traded
      */
-    function _checkApprovals(uint256 _uniAmount, uint256 _sushiAmount, ERC20 _token) internal {
+    function _checkApprovals(uint256 _uniAmount, uint256 _sushiAmount, IERC20 _token) internal {
         if (_token.allowance(address(this), address(uniRouter)) < _uniAmount) {
-            _token.approve(address(uniRouter), PreciseUnitMath.MAX_UINT_256);
+            _token.approve(address(uniRouter), 2 ** 96 - 1);
```
bweick, line 303: I don't think we want to do this. The unlimited approve flow before is probably fine, was just noting that there may be some tokens that this could fail for if they don't take that into account (though it is a common heuristic so I would imagine a lot of them do)_

contracts/product/AMMSplitter.sol
```diff
@@ -204,51 +205,52 @@ contract AMMSplitter {
     }
 
     /**
-     * Calculates the optimal trade sizes for Uniswap and Sushiswap.
+     * Calculates the optimal trade sizes for Uniswap and Sushiswap. Pool vaules must be measured in the same token. For single hop trades
```
bweick, line 208: Really good javadoc here, this is great

contracts/product/AMMSplitter.sol
```diff
@@ -294,15 +296,29 @@ contract AMMSplitter {
...
+    /**
+     * Gets the balance of a component token in a Uniswap / Sushiswap pool
+     *
+     * @param _factory          factory contract to use (either uniFactory or sushiFactory)
+     * @param _tokenA           first token in pair
+     * @param _balanceToken     second token in pair, and token to get balance of
+     *
+     * @return uint256          balance of second token in pair
+     */
+    function _getTokenBalanceInPair(IUniswapV2Factory _factory, address _tokenA, address _balanceToken) internal view returns (uint256) {
```
bweick, line 319: Instead of `_tokenA` maybe just use `_pairedToken` or something to more explicitly indicate this is the other token in the pair

test/product/ammSplitter.spec.ts
```diff
@@ -612,6 +612,95 @@ describe("AMMSplitter", async () => {
           expect(finalTraderWbtc.sub(initTraderWbtc)).to.eq(expectedTotalOutput);
         });
       });
+
+      context("when a token with less than 18 decimals is the intermediary trade token", async () => {
```
bweick, line 616: Just curious did you run this test before the fix?

contracts/product/AMMSplitter.sol
```diff
@@ -0,0 +1,386 @@
...
+     * Tu/Ts = ((Psa + Psb) * Pua * Pub) / ((Pua + Pub) * Psa * Psb)
+     *
+     * Ts  = Sushiswap trade size
+     * Tu  = Uniswap trade size
+     * Pua = Uniswap liquidity for pool A
+     * Pub = Uniswap liquidity for pool B
+     * Psa = Sushiswap liquidity for pool A
+     * Psb = Sushiswap liquidity for pool B
+     *
+     * @param _uniValueA        Size of the first Uniswap pool
```
bweick, line 269: Realizing these variable names may not be the best, maybe a pattern like `sushiLiqPoolA` is better, describes what exactly the value is. Value is just too generic. I'd apply that to `getTradeSizes` as well

contracts/product/AMMSplitter.sol
```diff
@@ -0,0 +1,386 @@
...
+        uint256 _uniValueA,
+        uint256 _uniValueB,
+        uint256 _sushiValueA,
+        uint256 _sushiValueB
+    ) 
+        internal
+        pure
+        returns (uint256)
+    {
+        uint256 a = _sushiValueA.add(_sushiValueB).preciseDiv(_uniValueA.add(_uniValueB));
```
bweick, line 286: Great job refactoring this, this looks like it should work

contracts/product/AMMSplitter.sol
```diff
@@ -0,0 +1,386 @@
...
+import { IUniswapV2Router } from "../interfaces/external/IUniswapV2Router.sol";
+import { PreciseUnitMath } from "../lib/PreciseUnitMath.sol";
+
+/**
+ * @title AMMSplitter
+ * @author Set Protocol
+ *
+ * Peripheral contract which splits trades efficiently between Uniswap V2 and Sushiswap. Works for both exact input 
+ * and exact output trades. All math for calculating the optimal split is performed on-chain. This contract only supports
+ * trade paths of up to two hops.
```
bweick, line 35: In your STIP you list some assumptions that go into the trade splitting logic (ie price impact is low percentage, pools have same price) can you list those assumptions here, think that's really pertinent info for reviewers.

test/product/ammSplitter.spec.ts
```diff
@@ -0,0 +1,1708 @@
...
+
+          await subject();
+
+          const finalTraderWbtc = await setup.wbtc.balanceOf(subjectTo.address);
+
+          expect(finalTraderWbtc.sub(initTraderWbtc)).to.eq(expectedTotalOutput);
+        });
+      });
+
+      context("when 70% of the liquidity is in the Uniswap pools", async () => {
```
bweick, line 418: These tests are in general really great one case that may be good to check on is when you do multi-hop and the B pool for Sushi is the biggest but the A pool for Uni is the biggest. Would be interested in seeing a test for that. Think the test will also give a good understanding to other reviewers of how it works

test/product/ammSplitter.spec.ts
ncitron: No this test is new

richardliang (commented):
Only reviewed the comments / javadocs. In general, looked for these 3 points:

```
Do not comment the obvious
Do not comment what the code is doing
Comment why you are doing it.
```

https://francescolelli.info/programming/best-practices-on-how-to-write-comments-in-your-code/

bweick (approved):

richardliang (approved):

PR #101, opened 2021-06-21
Add Kyber V3 Index Exchange adapter

Alpha-Guy:
**Why?** 
Because KNC kills us during rebalancing. The liquidity for KNC/ETH is very thin on Uniswap V2 and thus we have to do 30 trades on average to rebalance KNC token in DPI. 

**Solution**
Move to using Kyber's latest [DMM Exchange](https://dmm.exchange) as the source of our liquidity for KNC/ETH pair.

**Liquidity for KNC/ETH pair**
- Uniswap V3 = No liquidity available
- Sushiswap = No liquidity available
- Uniswap V2 = 1,749,990
- Kyber V3 DMM exchange = 3,848,968
   - Effective liqudity on DMM = Total liquidity in pool * Amp factor = 3,848,968 * 1.9 = 7,313,039

**Ratio of Kyber V3 liquidity to Uniswap V2 Liquidity = 7,313,039 / 1,749,990 = 4.1789**

### Implementation
- The KyberV3 DMM contract interfaces are similar to Uniswap V2 contracts. It has a router contract with our familiar swap functions, `swapExactTokensForTokens` and `swapTokensForExactTokens`.

**Important Differences:**
- Each pool in Kyber V3 has an amplification factor which makes the pool trade only in a specified price range and within a given region of the constant product function. Thus making the pools more capital efficient.
- For each token pair, there can be possibly many pools with different configurations for the pricing curve (i.e different amplification factors). Hence the address of the pool to be used for trading must be passed by the off-chain infrastructure.
    - Managers can refer to [DMM Analytics](https://info.dmm.exchange/) page to select the pool with the highest liquidity.
    - Off-chain scripts can use [instructions given here](https://docs.dmm.exchange/developer-guides/smart-contract-integration/fetching-pool-addresses/#selecting-the-right-pool) to select the pool with the highest liquidity.

*Relevant parts of Kyber docs:*
- [Programmable pricing curve](https://docs.dmm.exchange/programmable-pricing-curve)
- [Swap Execution](https://docs.dmm.exchange/developer-guides/smart-contract-integration/swap-execution)
- [swapTokensForExactTokens](https://docs.dmm.exchange/reference/smart-contract/dmmRouter02#swaptokensforexacttokens)
- [swapExactTokensForTokens](https://docs.dmm.exchange/reference/smart-contract/dmmRouter02#swapexacttokensfortokens)

### Progress
- [x] Add `KyberV3IndexExchangeAdapter` contract
- [x] Write `KyberV3 DMM` fixtures
- [x] Write unit tests for adapter contract
- [x] Write unit tests for `Kyber V3 DMM` fixture
- [x] Write integration tests with GIM

**Deadline**: July Rebalance

### Blockages
- As DMM exchange has liquidity only in the new KNC tokens, we would first need to migrate DPI to new KNC tokens, using the existing `KyberMigrationAdapter`.


test/protocol/integration/index-exchange/kyberV3IndexExchangeAdapter.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+      });
+
+      it("should return the correct trade calldata", async () => {
+        const calldata = await subject();
+        const callTimestamp = await getLastBlockTimestamp();
+        const expectedCallData = kyberSetup.dmmRouter.interface.encodeFunctionData("swapTokensForExactTokens", [
+          destinationQuantity, // Source and destination quantity are flipped for swapTokensForExactTokens
+          sourceQuantity,
+          [poolAddress],
+          [sourceToken, destinationToken],
```
ncitron, line 160: Does this path need to have `sourceToken` and `destinationToken` flipped or does Kyber handle that?

test/protocol/modules/generalIndexModule.spec.ts
```diff
@@ -1125,6 +1166,112 @@ describe("GeneralIndexModule", () => {
...
+              const currentDaiAmount = await setup.dai.balanceOf(subjectSetToken.address);
+              const currentWethAmount = await setup.weth.balanceOf(subjectSetToken.address);
+              const totalSupply = await subjectSetToken.totalSupply();
+
+              await subject();
+
+              const lastBlockTimestamp = await getLastBlockTimestamp();
+
+              const expectedWethPositionUnits = preciseDiv(currentWethAmount.add(expectedOut), totalSupply);
+              const expectedDaiPositionUnits = preciseDiv(currentDaiAmount.sub(ether(1000)), totalSupply);
```
ncitron, line 1198: Can we make it more clear how 1000 is calculated as the sell amount. Maybe do something like
`const daiToSell = // calculation for the sell amount`
I recently had to write some tests for GIM, and had to figure out the math from scratch, and would have appreciated having the tests at least document the math (but better if it just does that math in a variable declaration as I showed above)

test/protocol/modules/generalIndexModule.spec.ts
```diff
@@ -1125,6 +1166,112 @@ describe("GeneralIndexModule", () => {
...
+                kyberV3ExchangeAdapter.address,
+                trader.address,
+                ether(1000),
+                expectedOut,
+                ZERO
+              );
+            });
+          });
+
+          describe("when component is being bought using UniswapV3", async () => {
```
ncitron, line 1224: KyberV3

test/protocol/modules/generalIndexModule.spec.ts
```diff
@@ -1125,6 +1166,112 @@ describe("GeneralIndexModule", () => {
...
+              subjectComponent = setup.wbtc.address;
+              subjectEthQuantityLimit = MAX_UINT_256;
+            });
+
+            afterEach(async () => {
+              await indexModule.setExchanges(subjectSetToken.address, [setup.wbtc.address], [sushiswapAdapterName]);
+            });
+
+            it("the position units and lastTradeTimestamp should be set as expected", async () => {
+              const amountOut = bitcoin(0.1);
```
ncitron, line 1240: Same comment as before about adding some clarity about how you calculate the 0.1 wBTC sell amount

utils/fixtures/kyberV3DMMFixture.ts
```diff
@@ -0,0 +1,67 @@
...
+      BigNumber.from(10000)   // Amp factor of 1 (in BPS) because un-correlated assets
+    );
+    this.wethWbtcPool = await this.createNewPool(
+      _weth,
+      _wbtc,
+      BigNumber.from(15000)   // Amp factor of 1.5 (in BPS) because correlated assets
+    );
+  }
+
+  public async createNewPool(_tokenA: Address, _tokenB: Address, _ampBps: BigNumber): Promise<DMMPool> {
```
ncitron, line 58: Does this fail when the token order is wrong? If so, it would be nice if you could bake in a call to `getTokenOrder` so that users of the fixture don't have to worry about it

ncitron (commented):
Overall looks really good. Left a couple of small questions and comments

test/protocol/modules/generalIndexModule.spec.ts
```diff
@@ -11,7 +11,9 @@ import {
   GeneralIndexModule,
```
ncitron, line 11: Also, we've been trying to separate out integration tests for new exchange and index-exchange adapters into new files now to prevent the main test from becoming too big over time. Take a look out how I did it here:
https://github.com/SetProtocol/set-protocol-v2/pull/99/files#diff-d567ba1527ee362b53fe24be298c60b9fc43228d3ecfa1d4496196859d3ed4d3R1

contracts/protocol/integration/index-exchange/KyberV3IndexExchangeAdapter.sol
```diff
@@ -0,0 +1,132 @@
...
+        view
+        override
+        returns (address, uint256, bytes memory)
+    {
+        address[] memory path = new address[](2);
+        path[0] = _sourceToken;
+        path[1] = _destinationToken;
+
+        address[] memory poolsPath = new address[](1);
+        poolsPath[0] = _data.toAddress(0);
```
bweick, line 110: We should validate that this isn't a zero address. If there's any easy way to validate that it's a Kyber pool too that'd be awesome

contracts/protocol/integration/index-exchange/KyberV3IndexExchangeAdapter.sol
```diff
@@ -0,0 +1,132 @@
...
+
```
bweick, line 111: If it's easy we can also check to make sure the tokens are both in the passed pool (though not interested in adding too much gas)

test/protocol/integration/index-exchange/kyberV3IndexExchangeAdapter.spec.ts
bweick: Just the quantities need to be since the destinationQuantity -> fixedOutput and sourceQuantity goes to variableInput (which is the ordering in the function signature)

test/protocol/modules/generalIndexModule.spec.ts
bweick: +1 this would be really nice

test/protocol/integration/index-exchange/kyberV3IndexExchangeAdapter.spec.ts
Alpha-Guy: Kyber expects [`tokenIn`, `tokenOut`] as path. The GIM decides which is `tokenIn` and `tokenOut`, and passes them as `sourceToken` and `destinationToken` respectively. 
Quantities are flipped just because those two functions in Kyber router, except them in that order, the fixed amount as first argument and variable amount as second.

utils/fixtures/kyberV3DMMFixture.ts
Alpha-Guy: The token addresses are interchangeable. I added it in the javadocs.
```
  /**
   * Creates new DMM pool. The token addresses are interchangeable.
   * NOTE: There can be at most 1 unamplified pool for a token pair, ie. only 1 pool can exist with ampBps = BPS (10000). 
   * Should there already be an existing unamplified pool, attempts to create another one will fail.
   *
   * @param _token0     address of token 1
   * @param _token1     address of token 2
   * @param _ampBps     Amplification factor (in BPS)
   */
   ```

File: contracts/protocol/integration/index-exchange/KyberV3IndexExchangeAdapter.sol [hunk 9]
```diff
@@ -0,0 +1,138 @@
...
+        uint256 _sourceQuantity,
+        uint256 _destinationQuantity,
+        bytes memory _data
+    )
+        external
+        view
+        override
+        returns (address, uint256, bytes memory)
+    {
+        require(dmmFactory.isPool(IERC20(_sourceToken), IERC20(_destinationToken), _data.toAddress(0)), "Invalid pool address");
```
bweick, line 109: Let's move this validation down after line 116. Don't need to decode that data twice

bweick (approved):

PR #110, opened 2021-07-09
Update UniswapV2IndexExchangeAdapter to support two-hop trades

ncitron:
### Why
Currently, all index exchange adapters only support single hop trades. With the addition of Badger to the DPI, we will need to execute two-hop trade when rebalancing, since Badger is most liquid with wBTC.

### Solution
Update `UniswapV2IndexExchangeAdapter` to support multi-hop trades. The Badger token is most liquid on Sushiswap, so this adapter will be sufficient to fill our needs.

### Implementation
- In `getTradeCalldata` check if the _data parameter is empty. If so, create the trade path using the input and output tokens (single-hop trade)
- If _data is not empty, then interpret it as a single address. This address will be the intermediate token of the two-hop trade

### Core Differences
- For `GeneralIndexModule` to perform a two-hop trade, ensure that you use `setExchangeData` beforehand to encode the intermediate token

### DIFF
--- a/contracts/protocol/integration/index-exchange/UniswapV2IndexExchangeAdapter.sol
+++ b/contracts/protocol/integration/index-exchange/UniswapV2IndexExchangeAdapter.sol
@@ -21,6 +21,7 @@ pragma experimental "ABIEncoderV2";
 
 import { IIndexExchangeAdapter } from "../../../interfaces/IIndexExchangeAdapter.sol";
 
+
 /**
  * @title UniswapV2IndexExchangeAdapter
  * @author Set Protocol
@@ -68,7 +69,8 @@ contract UniswapV2IndexExchangeAdapter is IIndexExchangeAdapter {
      * @param  _isSendTokenFixed         Boolean indicating if the send quantity is fixed, used to determine correct trade interface
      * @param  _sourceQuantity           Fixed/Max amount of source token to sell
      * @param  _destinationQuantity      Min/Fixed amount of destination token to buy
-     * @param  _data                     Encoded address intermediary token in the trade path. If empty, path is the input and output tokens
+     * @param  _data                     Encoded address intermediary token in the trade path. If empty, path is the input and output tokens. 
+                                         Note: only allows one intermediary asset
      *
      * @return address                   Target contract address
      * @return uint256                   Call value
--- a/contracts/protocol/integration/index-exchange/UniswapV2IndexExchangeAdapter.sol
+++ b/contracts/protocol/integration/index-exchange/UniswapV2IndexExchangeAdapter.sol
@@ -27,6 +27,9 @@ import { IIndexExchangeAdapter } from "../../../interfaces/IIndexExchangeAdapter
  *
  * A Uniswap Router02 exchange adapter that returns calldata for trading with GeneralIndexModule, allows encoding a trade with a fixed input quantity or
  * a fixed output quantity.
+ *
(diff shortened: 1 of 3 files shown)


contracts/protocol/integration/index-exchange/UniswapV2IndexExchangeAdapter.sol
```diff
@@ -65,6 +68,7 @@ contract UniswapV2IndexExchangeAdapter is IIndexExchangeAdapter {
      * @param  _isSendTokenFixed         Boolean indicating if the send quantity is fixed, used to determine correct trade interface
      * @param  _sourceQuantity           Fixed/Max amount of source token to sell
      * @param  _destinationQuantity      Min/Fixed amount of destination token to buy
+     * @param  _data                     Encoded address intermediary token in the trade path. If empty, path is the input and output tokens
```
bweick, line 71: Think it'd be good to make a note that this will only work for one intermediate asset

contracts/protocol/integration/index-exchange/UniswapV2IndexExchangeAdapter.sol
```diff
@@ -77,16 +81,26 @@ contract UniswapV2IndexExchangeAdapter is IIndexExchangeAdapter {
...
-        path[0] = _sourceToken;
-        path[1] = _destinationToken;
+        address[] memory path;
+
+        if (_data.length == 0) {
+            path = new address[](2);
+            path[0] = _sourceToken;
+            path[1] = _destinationToken;
+        } else {
+            address intermediateToken = abi.decode(_data, (address));
```
bweick, line 98: Validate that the zero address isn't passed in

contracts/protocol/integration/index-exchange/UniswapV2IndexExchangeAdapter.sol
```diff
@@ -77,16 +81,26 @@ contract UniswapV2IndexExchangeAdapter is IIndexExchangeAdapter {
...
         view
         override
         returns (address, uint256, bytes memory)
     {
-        address[] memory path = new address[](2);
-        path[0] = _sourceToken;
-        path[1] = _destinationToken;
+        address[] memory path;
+
+        if (_data.length == 0) {
```
bweick, line 93: Does it make sense to check that the length is 0 or that the length is equal to an address's length? For example, right now we actually store 1 byte of data because etherscan has trouble rendering the executionInfo on the GIM when no data is specified. Kind've a weird niche case but wondering if that's worth still preserving.

bweick: What do you think is the ideal way this is done?

ncitron: Well an ETH address is 20 bytes, so we can just do something like:
```solidity
if (_data.length == 0) {
  // single hop
} else if (_data.length == 20) {
  // two-hop
} else {
  // revert
}
```

bweick: Ok I'm fine with this structure we'll figure a way around it, also optimistic we could get Etherscan to make some fixes. I'd prefer the security of being explicit that length 0 bytes means no data is supposed to be passed

bweick (approved):

PR #125, opened 2021-08-18
Add Debt Issuance Module V2

Alpha-Guy:
STIP: https://github.com/SetProtocol/STIPS/pull/8

### DIFF
--- a/contracts/protocol/lib/IssuanceUtils.sol
+++ b/contracts/protocol/lib/IssuanceUtils.sol
@@ -42,16 +42,14 @@ library IssuanceUtils {
      *
      * @param _setToken             Instance of the SetToken being issued/redeemed
      * @param _component            Address of component being transferred in/out
+     * @param _initialSetSupply     Initial SetToken supply before issuance/redemption
      * @param _componentQuantity    Amount of component transferred into SetToken
-     * @param _isIssue              True if issuing SetToken, false if redeeming
-     * @param _setQuantity          Amount of SetToken burnt in this transaction. If being issued, pass 0.
      */
     function validateCollateralizationPostTransferInPreHook(
         ISetToken _setToken, 
         address _component, 
-        uint256 _componentQuantity, 
-        bool _isIssue, 
-        uint256 _setQuantity
+        uint256 _initialSetSupply,
+        uint256 _componentQuantity
     )
         internal
         view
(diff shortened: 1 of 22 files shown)


contracts/protocol/modules/AaveDebtIssuanceModule.sol
```diff
@@ -0,0 +1,117 @@
...
+
+import { DebtIssuanceModule } from "./DebtIssuanceModule.sol";
+import { IController } from "../../interfaces/IController.sol";
+import { Invoke } from "../lib/Invoke.sol";
+import { ISetToken } from "../../interfaces/ISetToken.sol";
+import { PreciseUnitMath } from "../../lib/PreciseUnitMath.sol";
+
+
+/**
+ * @title AaveDebtIssuanceModule
```
bweick, line 35: In theory this can be used for many more tokens than just aTokens. We should name this in a way that captures the difference in collateralization assumptions post issue and redeem not something specific to aTokens.

bweick:
Fix coverage not passing

contracts/protocol/modules/DebtIssuanceModuleV2.sol
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * @title DebtIssuanceModuleV2
+ * @author Set Protocol
+ *
+ * The DebtIssuanceModuleV2 is a module that enables users to issue and redeem SetTokens that contain default and all
+ * external positions, including debt positions. Module hooks are added to allow for syncing of positions, and component
+ * level hooks are added to ensure positions are replicated correctly. The manager can define arbitrary issuance logic
+ * in the manager hook, as well as specify issue and redeem fees.
+ * 
+ * NOTE: This module supports issuing/redeeming SetTokens which hold one or more aTokens as components.
```
bweick, line 40: This doesn't provide sufficient context around this modules existence. This may allow us to mint/redeem Sets with aTokens, and that's worth noting as a practical manner. But it's critical we highlight the different collateralization logic. We need a thorough explanation that this module explicitly checks for collateralization of Default positions so that developers understand that core difference.

contracts/protocol/modules/DebtIssuanceModuleV2.sol
```diff
@@ -0,0 +1,95 @@
...
+            uint256 componentQuantity = _componentEquityQuantities[i];
+            if (componentQuantity > 0) {
+                if (_isIssue) {
+                    SafeERC20.safeTransferFrom(
+                        IERC20(component),
+                        msg.sender,
+                        address(_setToken),
+                        componentQuantity
+                    );
+                    IssuanceUtils.validateComponentTransfer(_setToken, component, _isIssue, _quantity);
```
bweick, line 82: Realizing this validation only works for `Default` positions...consider an External position where we want to check enough has been transferred in prior to sending it to another protocol. Since `validateComponentTransfer` only looks at the `Default` position for a component if there exists an `External` position we would expect the sum of `totalSupply*defaultPositionUnit + _quantity*externalPositionUnit` to be held by the Set.

contracts/protocol/lib/IssuanceUtils.sol
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Validates the component token transfer to/from SetToken during issuance/redemption. Reverts if Set is undercollateralized post transfer.
+     *
+     * @param _setToken         Instance of the SetToken being issued/redeemed
+     * @param _component        Address of component being transferred in/out
+     * @param _isIssue          True if issuing SetToken, false if redeeming
+     * @param _issueQuantity    Total SetToken issue quantity with fees. Pass 0 if Set is being redeemed.
+     */
+    function validateComponentTransfer(ISetToken _setToken, address _component, bool _isIssue, uint256 _issueQuantity) internal {
```
bweick, line 48: Per comment in `DebtIssuanceModuleV2` this needs to also take into account any `External` position(s) that may exist for the passed component. This interface probably needs to be reworked in order to accomodate that

bweick (approved):

PR #131, opened 2021-08-27
Upgrade hardhat and ethers (london hf)

### LINKED ISSUE #140: Upgrade to Typescript 4.x
I attempted to upgrade ethers in this repo (see [branch](https://github.com/rootulp/set-protocol-v2/tree/rp/upgrade-ethers)) but it looks like upgrading Typescript to 4.x is a prerequisite and can be separated from a distinct ethers upgrade PR

cgewecke:
Closes #140, #141  

PR  
+ upgrades Hardhat and associated deps to the London HF 
+ upgrade ethers to latest for EIP1559 support
+ remove typechain file re-writing logic for installation and test commands
+ updates library link id generation for new format in `typechain/hardhat`
+ fixes stray compiler warning in a mock solidity file.

**Companion PRs**

Have tested these changes by publishing an `rc` version of this branch to npm and installing in our downstream repos:
+ [set.js PR 76][11]
+ [set-v2-deployments PR 123][12]

[11]: https://github.com/SetProtocol/set.js/pull/76
[12]: https://github.com/SetProtocol/set-v2-deployments/pull/123 

**Added**
+ @typechain/hardhat (hardhat-typechain repo has migrated here)

**Upgrades**
+ hardhat -> v2.6.1
+ solidity-coverage -> v0.7.17
+ ethers -> 5.4.6
+ ethereum-waffle -> 3.4.0
+ typescript -> v4
+ typechain -> 5.1.2
+ @typechain/ethers-v5 -> 7.0.1
+ CircleCI Docker containers to Node 12

**Removed**: these are unused or replaced
+ @typechain/truffle-4,
+ @typechain/truffle-5
+ @typechain/web3-1
+ hardhat-deploy 
+ hardhat-typechain (replaced with/ @typechain/hardhat)

**Additional**
+ Rewrite all vyper-compiled contract abi gas fields to be string instead of number (for typechain). 
+ Remove fix-typechain.js and misc typechain hack yarn commands


cgewecke:
Forked tests failing here because:

> The server is temporarily unable to service your
request due to maintenance downtime or capacity
problems. Please try again later.

(Have updated the alchemy secret in CI to use the Set api key)

package.json
```diff
@@ -52,15 +49,19 @@
   "author": "felix2feng",
   "license": "MIT",
   "homepage": "https://github.com/SetProtocol",
+  "resolutions": {
+    "@ethersproject/abi": "5.4.1",
+    "@ethersproject/bignumber": "5.4.1",
+    "@ethersproject/contracts": "5.4.1",
+    "@ethersproject/providers": "5.4.5"
+  },
```
cgewecke, line 57: These "resolutions" are necessary to coerce ethers monorepo packages to the right version for ethers `5.4.6`. 

There are several versions of ethers (and its sub-packages) nested in the dependency tree and they don't have consistent type definitions. We import stuff directly from sub-packages all over the place, so there's a lot of version mis-matching and tsc compilation blood.

bweick: If we started from scratch is there some other way you would do this? (i.e. import some of these dependencies differently in various files)

bweick: Not suggesting we do that, more for my edification.

bweick (approved):

package.json
cgewecke: Yes, good question. I guess I would rewrite all the `import { something } from @ethersproject/xyz` to pull from `ethers` directly, instead of a sub-package. 

The problem is that if we have a dependency like ethereum-waffle and it has `@ethersproject/xyz@5.0.12`, yarn will install as below:
```
@ethersproject /
  xyz@5.0.12
ethers/
  node_modules/
    @ethersproject/
      xyz@5.4.6
```
... and when we import `@ethersproject/xyz` we get the outermost (ethereum-waffle) one. Its types confilct with whatever typechain generated or something else we're using from the main ethers package.

PR #134, opened 2021-08-31
ALM integration tests

Alpha-Guy:
- [x] Fix flaky tests and make them pass every time
- [x] Explain reasons for using WBTC
- [x] In Issuance, add test for the case when default and external positions overlap
- [x] In redemption, add tests for the case when 2 default positions and 2 external positions
    - [x] Test both overlapping and non-overlapping cases
- [x] Avoid using `greater than or equal to` comparisons.
- [x] When can't avoid `gte` comparisons. Replace it with `greater than or equal to with an upper bound` comparison. And give a clear reason to use it.

### DIFF
--- a/test/integration/aaveUniswapLeverageDebtIssuance.spec.ts
+++ b/test/integration/aaveUniswapLeverageDebtIssuance.spec.ts
@@ -37,7 +37,7 @@ import { ADDRESS_ZERO, ZERO, EMPTY_BYTES, MAX_UINT_256 } from "@utils/constants"
 
 const expect = getWaffleExpect();
 
-describe.skip("AaveUniswapLeverageDebtIssuance", () => {
+describe("AaveUniswapLeverageDebtIssuance", () => {
   let owner: Account;
   let feeRecipient: Account;
   let deployer: DeployHelper;
@@ -50,7 +50,7 @@ describe.skip("AaveUniswapLeverageDebtIssuance", () => {
   let debtIssuanceModule: DebtIssuanceModuleV2;
   let uniswapExchangeAdapter: UniswapV2ExchangeAdapter;
 
-  let aWETH: AaveV2AToken;
+  let aWBTC: AaveV2AToken;
   let variableDebtDAI: AaveV2VariableDebtToken;
   let variableDebtUSDC: AaveV2VariableDebtToken;
 
@@ -71,6 +71,18 @@ describe.skip("AaveUniswapLeverageDebtIssuance", () => {
     aaveV2Setup = getAaveV2Fixture(owner.address);
     await aaveV2Setup.initialize(setup.weth.address, setup.dai.address);
 
+
+    // Create a WBTC reserve
+    const wbtcReserveTokens = await aaveV2Setup.createAndEnableReserve(
+      setup.wbtc.address, "WBTC", BigNumber.from(8),
+      BigNumber.from(8000),   // base LTV: 80%
+      BigNumber.from(8250),   // liquidation threshold: 82.5%
+      BigNumber.from(10500),  // liquidation bonus: 105.00%
+      BigNumber.from(1000),   // reserve factor: 10%
+      true,                   // enable borrowing on reserve
+      true                    // enable stable debts
+    );
+
     // Create an USDC reserve
     const usdcReserveTokens = await aaveV2Setup.createAndEnableReserve(
       setup.usdc.address, "USDC", BigNumber.from(6),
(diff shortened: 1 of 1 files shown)


test/integration/aaveUniswapLeverageDebtIssuance.spec.ts
```diff
@@ -541,203 +537,347 @@ describe.skip("AaveUniswapLeverageDebtIssuance", () => {
...
+          const postSetUsdcDebtBalance = await variableDebtUSDC.balanceOf(subjectSetToken);
+          const postMinterDaiBalance = await setup.dai.balanceOf(subjectCaller.address);
+          const postSetDaiDebtBalance = await variableDebtDAI.balanceOf(subjectSetToken);
+          const postSetUsdcEquityBalance = await setup.usdc.balanceOf(subjectSetToken);
+
+          expect(postMinterAWBTCBalance).to.eq(preMinterAWBTCBalance.sub(aWBTCFlows));
+          expect(postSetAWBTCBalance).to.eq(preSetAWBTCBalance.add(aWBTCFlows));
+          expect(postMinterUsdcBalance).to.eq(preMinterUsdcBalance.sub(usdcEquityFlows).add(usdcDebtFlows));
+          expect(postSetUsdcEquityBalance).to.eq(preSetUsdcEquityBalance.add(usdcEquityFlows));
+          expect(postSetUsdcDebtBalance).to.gte(preSetUsdcDebtBalance.add(usdcDebtFlows)); // Round up due to interest accrual
```
bweick, line 869: Don't love this test cause the case we're trying to test for is one where we accidentally lever up twice, which would not be caught here

test/integration/aaveUniswapLeverageDebtIssuance.spec.ts
```diff
@@ -1208,7 +1322,7 @@ describe.skip("AaveUniswapLeverageDebtIssuance", () => {
         const initialPositions = await setToken.getPositions();
 
         await subject();
-        // aWETH position is increased
+        // aWBTC position is increased
```
bweick, line 1325: We should have a test to make sure we don't repay extra borrows since that uses a different hook

bweick (commented):
Can you expand on the reasoning why aBTC doesn't face some of the same rounding issues as aETH? Where were you getting reverts for aETH? Is it possible we just need to issue more Sets?

Other than that a couple comments. I think we need to beef up the testing on the double borrow/repay stuff.

test/integration/aaveUniswapLeverageDebtIssuance.spec.ts
Alpha-Guy: It would be caught here. When we accidentally lever up twice, the above token balance checks would fail and instead
```javascript
expect(postSetUsdcDebtBalance).to.gte(preSetUsdcDebtBalance.add(usdcDebtFlows.mul(2)));
```
 would pass.

test/integration/aaveUniswapLeverageDebtIssuance.spec.ts
```diff
@@ -711,11 +725,13 @@ describe("AaveUniswapLeverageDebtIssuance", () => {
...
           expect(postSetAWBTCBalance).to.eq(preSetAWBTCBalance.add(aWBTCFlows));
           expect(postMinterUsdcBalance).to.eq(preMinterUsdcBalance.add(usdcFlows));
-          expect(postSetUsdcDebtBalance).to.gte(preSetUsdcDebtBalance.add(usdcFlows)); // Round up due to interest accrual
           expect(postMinterDaiBalance).to.eq(preMinterDaiBalance.add(daiFlows));
-          expect(postSetDaiDebtBalance).to.gte(preSetDaiDebtBalance.add(daiFlows)); // Round up due to interest accrual
           expect(postMinterWethBalance).to.eq(preMinterWethBalance.sub(wethFlows));
           expect(postSetWethBalance).to.eq(preSetWethBalance.add(wethFlows));
+          // USDC has 6 decimal places. Interest accrued doesn't reflect in balance immediately.
+          expect(postSetUsdcDebtBalance).to.eq(preSetUsdcDebtBalance.add(usdcFlows));
+          // DAI has 18 decimal places. Thus we need to use greater than or equal to here to account for the very small amount of interest accrued.
```
bweick: Still think we should add a check that makes sure it's between .9 and 1.1 debt units for tokens that have 18 decimals. That said great explanations otherwise, makes sense why BTC and USDC works and Dai we can't be as exact.

bweick (approved):

Issue #106, opened 2021-06-29
Error HH8: There's one or more errors in your config file:


skolaautomat:
Hi, I downloaded the code and tried to compile it but I experience many issues in the default configuration for developers:

Error HH8: There's one or more errors in your config file:

  * Invalid value {"url":"https://kovan.infura.io/v3/cccccccccccccccccccccccccccccccc","accounts":["0xundefined"]} for HardhatConfig.networks.kovan - Expected a value of type HttpNetworkConfig.
  * Invalid value {"url":"https://mainnet.infura.io/v3/cccccccccccccccccccccccccccccccc","accounts":["0xundefined"]} for HardhatConfig.networks.staging_mainnet - Expected a value of type HttpNetworkConfig.
  * Invalid value {"url":"https://mainnet.infura.io/v3/cccccccccccccccccccccccccccccccc","accounts":["0xundefined"]} for HardhatConfig.networks.production - Expected a value of type HttpNetworkConfig.
  
To learn more about Hardhat's configuration, please go to https://hardhat.org/config/

For more info go to https://hardhat.org/HH8 or run Hardhat with --show-stack-traces
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

viratjshah:
I am facing same issue

shankinson:
@skolaautomat @viratjshah I was just able to get things compiled and tests passing. I did the following:

For Mac, install solidity 0.6.10 as per https://docs.soliditylang.org/en/v0.8.7/installing-solidity.html, git hash taken from the Set Protocol Readme (https://github.com/SetProtocol/set-protocol-v2/blob/master/README.md)

```
brew update
brew upgrade
brew tap ethereum/ethereum
brew install solidity
git clone https://github.com/ethereum/homebrew-ethereum.git
cd homebrew-ethereum
git checkout 06d13a8499801dc3ea4f19b2d24ed2eeb3072ebb
brew unlink solidity
brew install solidity.rb
```

Then I essentially followed the steps shown in the circleci yaml file (https://github.com/SetProtocol/set-protocol-v2/blob/master/.circleci/config.yml):

```
git clone https://github.com/SetProtocol/set-protocol-v2.git
cd set-protocol-v2
cp .env.default .env
yarn install
yarn build
yarn chain # in a different terminal window
yarn test
yarn coverage
```

PR #136, opened 2021-09-07
Add getting started section to README

rootulp:
Resolves https://github.com/SetProtocol/set-protocol-v2/issues/106

I needed to do these three steps to be able to run `yarn` in this package. Without these steps, I encountered:

```
$ npx hardhat compile
Error HH8: There's one or more errors in your config file:

  * Invalid value {"url":"https://kovan.infura.io/v3/undefined","accounts":["0xundefined"]} for HardhatConfig.networks.kovan - Expected a value of type HttpNetworkConfig.
  * Invalid value {"url":"https://mainnet.infura.io/v3/undefined","accounts":["0xundefined"]} for HardhatConfig.networks.staging_mainnet - Expected a value of type HttpNetworkConfig.
  * Invalid value {"url":"https://mainnet.infura.io/v3/undefined","accounts":["0xundefined"]} for HardhatConfig.networks.production - Expected a value of type HttpNetworkConfig.

To learn more about Hardhat's configuration, please go to https://hardhat.org/config/
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,14 +15,20 @@
 ### Getting started
 
 1. Clone this repo
-2. Navigate to [Alchemy](https://www.alchemy.com/) and create an application
-3. Copy the default environment variables file and populate `ALCHEMY_TOKEN` with the token generated in the previous step
+1. Create an Alchemy token
+    1. Navigate to [Alchemy](https://www.alchemy.com/) and create an account
+    1. Click **Create app** and provide a name (e.g. set-protocol-v2)
+    1. Navigate to your new app and click **View key** then copy the token (the part after `https://eth-mainnet.alchemyapi.io/v2/`)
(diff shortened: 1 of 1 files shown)


bweick:
This is good! One thing I'd add though is the error you got there seems to be related to a bad `infura` value? Alchemy is required for mainnet fork testing. 

Did putting in an Alchemy key help solve this error? Maybe we can expand the steps to include getting an `infura` key?

rootulp:
> Did putting in an Alchemy key help solve this error?

Yep. Although agreed with your call out that the error explicitly states Infura so I'm not sure why my `.env` without a valid Infura token no longer encounters this error. I definitely don't mind adding steps for Infura as well. Will follow up with another commit

PR #144, opened 2021-09-10
Replace @ethersproject imports with ethers

rootulp:
Motivation in [this comment](https://github.com/SetProtocol/set-protocol-v2/pull/131#discussion_r705766240). I expect we'll encounter fewer dependency mismatches for ethers subpackages if we import from the umbrella package. Also the [ethers docs](https://github.com/ethers-io/ethers.js/tree/master/packages/bignumber) say:

> Most users will prefer to use the umbrella package, but for those with more specific needs, individual components can be imported.

### Testing
`yarn build` succeeds

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,12 +49,6 @@
   "author": "felix2feng",
   "license": "MIT",
   "homepage": "https://github.com/SetProtocol",
-  "resolutions": {
-    "@ethersproject/abi": "5.4.1",
-    "@ethersproject/bignumber": "5.4.1",
-    "@ethersproject/contracts": "5.4.1",
-    "@ethersproject/providers": "5.4.5"
-  },
   "devDependencies": {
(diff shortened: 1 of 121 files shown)


cgewecke:
Wow!! Will check this out...

cgewecke (commented):
This is super cool. 

The only additional thing to do here is remove the resolutions section in `package.json`.  

https://github.com/SetProtocol/set-protocol-v2/blob/59c819c7c6260e9ea2ebdc6b72d1e353df6d5f66/package.json#L52-L57

(Tried this locally and the build was clean)  You just need to 
+ delete the resolutions
+ run `rm -rf node_modules`
+ run yarn

...and the yarn lock should update appropriately.

cgewecke (approved):
Looks great! 

(Added a couple additional changes in one test file that uses ethereum-waffle mock contract fixtures)

PR #145, opened 2021-09-11
BalancerV2IndexExchangeAdapter

ncitron:


utils/fixtures/balancerV2Fixture.ts
```diff
@@ -0,0 +1,92 @@
+import { Signer } from "@ethersproject/abstract-signer";
+import { JsonRpcProvider, Web3Provider } from "@ethersproject/providers";
```
rootulp, line 2: [nit] if we import from the umbrella package `ethers` rather than `@ethersproject/foo`, we can avoid issues with mismatching dependencies for ethers sub-packages. Additional context in: https://github.com/SetProtocol/set-protocol-v2/pull/144 

Concretely, this means replacing this import with
```suggestion
import { providers } from "ethers";
```

and below

```typescript
  constructor(provider: providers.Web3Provider | providers.JsonRpcProvider, ownerAddress: Address) {
```

utils/fixtures/balancerV2Fixture.ts
```diff
@@ -0,0 +1,92 @@
...
+import { ether } from "@utils/common";
+import DeployHelper from "@utils/deploys";
+import { Account } from "@utils/test/types";
+import { Address } from "@utils/types";
+
+import { StandardTokenMock, WETH9 } from "../contracts/index";
+import { BVault, WeightedPoolFactory } from "../contracts/balancerV2";
+import { BPoolV2__factory } from "../../typechain/factories/BPoolV2__factory";
+import { BigNumber } from "@ethersproject/bignumber";
```
rootulp, line 11:
```suggestion
import { BigNumber } from "ethers";
```

utils/fixtures/balancerV2Fixture.ts
```diff
@@ -0,0 +1,92 @@
...
+import { defaultAbiCoder } from "@ethersproject/abi";
```
rootulp, line 12:
```suggestion
import { utils } from "@ethersproject/abi";
```

and below:

```typescript
    const initUserData = utils.defaultAbiCoder.encode(["uint256", "uint256[]"], [0, tokenAndAmounts.map(ta => ta.amount)]);
```

external/contracts/balancer/v2/BVault.sol
```diff
@@ -0,0 +1,91 @@
...
+Privacy
+Security
+Status
+Docs
+Contact GitHub
+Pricing
+API
+Training
+Blog
+About
```
rootulp, line 91: I think this is an artifact from copy + paste. Should it be removed?
```suggestion

```

rootulp (commented):
Sorry I'm not familiar enough w/ Solidity to provide a helpful review there

utils/fixtures/balancerV2Fixture.ts
ncitron: Wouldn't I want to import this from `ethers` instead? Importing `utils` from the same `@ethersproject` library shouldn't have any meaningful effect. Am I missing something?

external/contracts/balancer/v2/BVault.sol
ncitron: Goo catch

utils/fixtures/balancerV2Fixture.ts
rootulp: Sorry that's my bad, you're absolutely right. I meant:

```suggestion
import { utils } from "ethers";
```

PR #155, opened 2021-10-21
Curve amm adapter

RedVeil:
# Code Review Processes
## New Feature Review

README Checks
- [x] README has proper context for the reviewer to understand what the code includes, any important design considerations, and areas to pay more attention to

Code Checks
- [x] Add explanatory comments. If there is complex code that requires specific context or understanding, note that in a comment
- [x] Remove unncessary comments. Any comments that do not add additional context, information, etc. should be removed
- [] Add javadocs.  -- how do i do this?
- [x] Scrub through the code for inconsistencies (e.g. removing extra spaces)
- [x] Ensure there are not any .onlys in spec files

Broader Considerations
- [x] Ensure variable, function and event naming is clear, consistent, and reflective for the scope of the code.
- [x] Consider if certain pieces of logic should be placed in a different library, module


contracts/protocol/integration/amm/CurveFactoryMetapoolAmmAdapter.sol
```diff
@@ -0,0 +1,182 @@
...
+        }
+        return _literal;
+    }
+
+    function _getTokenIndex(address _pool, address _token) internal view returns (uint256) {
+        address[2] memory underlying = metapoolFactory.get_coins(_pool);
+        for (uint256 i = 0; i < 2; i++) {
+            if (underlying[i] == _token) return i;
+        }
+        revert("token not in pool");
```
bweick, line 180: Don't think we have coverage on this revert

curveFactoryMetapoolAmmAdapter.md
```diff
@@ -0,0 +1,9 @@
+## Curve FactoryMetapool Amm Adapter
```
bweick, line 1: This is great context, lets put it in the javadocs in the adapter contract and remove this markdown file. We like to keep all the context in the contract itself.

contracts/protocol/integration/amm/CurveFactoryMetapoolAmmAdapter.sol
```diff
@@ -0,0 +1,182 @@
...
+pragma solidity 0.6.10;
+
+import { SafeCast } from "@openzeppelin/contracts/utils/SafeCast.sol";
+
+import { IAmmAdapter } from "../../../interfaces/IAmmAdapter.sol";
+import { IMetapoolFactory } from "../../../interfaces/external/IMetapoolFactory.sol";
+import { IMetaPoolZap } from "../../../interfaces/external/IMetaPoolZap.sol";
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+contract CurveFactoryMetapoolAmmAdapter is IAmmAdapter {
```
bweick, line 28: If we can add some of the javadocs and section headers you see [here](https://github.com/SetProtocol/set-protocol-v2/blob/master/contracts/protocol/integration/amm/UniswapV2AmmAdapter.sol) for consistency.

contracts/protocol/integration/amm/CurveFactoryMetapoolAmmAdapter.sol
```diff
@@ -0,0 +1,182 @@
...
+        address[] calldata _components,
+        uint256[] calldata _maxTokensIn,
+        uint256 _minLiquidity
+    )
+        external
+        view
+        override
+        returns (address, uint256, bytes memory)
+    {
+        require(_isValidPool(_pool, _components), "invalid pool");
```
bweick, line 55: Don't need this because this check comes standard on the `AmmModule`, see [here](https://github.com/SetProtocol/set-protocol-v2/blob/3704b0cc0341f7e5e6db92d004584a1c0ef4fb63/contracts/protocol/modules/AmmModule.sol#L411)

contracts/protocol/integration/amm/CurveFactoryMetapoolAmmAdapter.sol
bweick: Though this revert may not be necessary because of `_isValidPool` we know that the passed components exist in that Meta Pool otherwise Line 160 would resolve `false`

contracts/protocol/integration/amm/CurveFactoryMetapoolAmmAdapter.sol
```diff
@@ -0,0 +1,182 @@
...
+
+    function isValidPool(address _pool, address[] memory _components) external view override returns(bool) {
+        return _isValidPool(_pool, _components);
+    }
+
+    function _isValidPool(address _pool, address[] memory _components) internal view returns(bool) {
+        address[2] memory expectedTokens = metapoolFactory.get_coins(_pool);
+
+        for (uint256 i = 0; i < _components.length; i++) {
+            if ((_components[i] == expectedTokens[0] || _components[i] == expectedTokens[1]) == false) {
```
bweick, line 160: I think we can rewrite this as `!(_components[i] == expectedTokens[0] || _components[i] == expectedTokens[1]) == false`

bweick, line 160: I think we want to check that there are no duplicates in the `_components` array, right now this would pass if there was a duplicate that was one of the tokens in the AMM.

But since the `AmmModule` is not yet deployed I say we make that check in the `AmmModule` itself because I can't think of a scenario where an AMM would have "duplicate" components.

bweick: Would add that check [here](https://github.com/SetProtocol/set-protocol-v2/blob/3704b0cc0341f7e5e6db92d004584a1c0ef4fb63/contracts/protocol/modules/AmmModule.sol#L406) using `AddressArrayUtils.hasDuplicate` which can be found in `contracts/lib`

contracts/protocol/integration/amm/CurveFactoryMetapoolAmmAdapter.sol
```diff
@@ -0,0 +1,182 @@
...
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+contract CurveFactoryMetapoolAmmAdapter is IAmmAdapter {
+    using SafeCast for uint256;
+    using SafeCast for int256;
+
+    IMetapoolFactory public metapoolFactory;
+    address public constant THREE_CRV = address(0x6c3F90f043a72FA612cbac8115EE7e52BDe6E490);
+
+    string public constant ADD_LIQUIDITY = "add_liquidity(uint256[2],uint256,address)";
```
ncitron, line 35: You probably want to use the ITriPoolZap functions rather than the IMetaPoolZap ones. This function requires depositing a combination of the 3Pool token and the extra token. The ITriPoolZap interface lets you deposit any of the 3Pool's pools' underlying tokens and the extra token. Makes it a lot more usable.

contracts/protocol/integration/amm/CurveFactoryMetapoolAmmAdapter.sol
```diff
@@ -0,0 +1,182 @@
...
+import { IMetapoolFactory } from "../../../interfaces/external/IMetapoolFactory.sol";
+import { IMetaPoolZap } from "../../../interfaces/external/IMetaPoolZap.sol";
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+contract CurveFactoryMetapoolAmmAdapter is IAmmAdapter {
+    using SafeCast for uint256;
+    using SafeCast for int256;
+
+    IMetapoolFactory public metapoolFactory;
+    address public constant THREE_CRV = address(0x6c3F90f043a72FA612cbac8115EE7e52BDe6E490);
```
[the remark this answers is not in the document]
anthonymartin: I think this can be deleted actually because it's not used

contracts/protocol/integration/amm/CurveFactoryMetapoolAmmAdapter.sol
anthonymartin: @ncitron it turns out that this interface is specifically for the factory metapool and not the legacy metapools. Do you know if the TriPoolZap supports the factory metapools? we had issues with the TriPoolZap reverting without any reason string response, so we ended up switching focus to the factory metapool with the intention to revisit the Zap and legacy metapool implementation afterwards.

cgewecke:
Closed in favor of #253

PR #157, opened 2021-10-28
Update getRequiredComponentIssuanceUnits to use token balances to cal…

bweick:
…culate default position issue amounts instead of position units.

This PR changes the `getRequiredComponentIssuanceUnits` getter to use the current component *balances* when calculating the amount of tokens required to issue a `Default` position instead of the `Default positionUnit`. We do this in order to provide better estimates for AAVE tokens which accrue tokens every block and are synced on every issuance and redemption for our FLI products.

We are treating this as a more one-off solution for dealing with AAVE. This module is now very specific to Sets that sync their balances on every issuance and redemption. `getRequiredComponentIssuanceUnits` will over estimate any Default positions that DO NOT sync every issuance and redemption AND have extra accrued tokens. If syncing is done through the Airdrop module and the manager has enabled an Airdrop fee then this will over-estimate the amount of tokens needed for issuance.

# Code Review Processes
## New Feature Review
* Design doc is created and posted here: **Not used here**

README Checks
- [x] README has proper context for the reviewer to understand what the code includes, any important design considerations, and areas to pay more attention to

Code Checks
- [x] Add explanatory comments. If there is complex code that requires specific context or understanding, note that in a comment
- [x] Remove unncessary comments. Any comments that do not add additional context, information, etc. should be removed
- [x] Add javadocs. 
- [x] Scrub through the code for inconsistencies (e.g. removing extra spaces)
- [x] Ensure there are not any .onlys in spec files

Broader Considerations
- [x] Ensure variable, function and event naming is clear, consistent, and reflective for the scope of the code.
- [x] Consider if certain pieces of logic should be placed in a different library, module

### DIFF
--- a/contracts/protocol/modules/DebtIssuanceModule.sol
+++ b/contracts/protocol/modules/DebtIssuanceModule.sol
@@ -390,6 +390,7 @@ contract DebtIssuanceModule is ModuleBase, ReentrancyGuard {
     )
         external
         view
+        virtual
         returns (address[] memory, uint256[] memory, uint256[] memory)
     {
         (
--- a/contracts/protocol/modules/DebtIssuanceModuleV2.sol
+++ b/contracts/protocol/modules/DebtIssuanceModuleV2.sol
@@ -27,6 +27,7 @@ import { IController } from "../../interfaces/IController.sol";
 import { Invoke } from "../lib/Invoke.sol";
 import { ISetToken } from "../../interfaces/ISetToken.sol";
 import { IssuanceValidationUtils } from "../lib/IssuanceValidationUtils.sol";
+import { Position } from "../lib/Position.sol";
 
 /**
  * @title DebtIssuanceModuleV2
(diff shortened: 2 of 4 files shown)


cgewecke (approved):
This looks good to me! Make sense.

To summarize, the main changes are:

+ how `cumulativeEquity` is calculated on L362 

```solidity
int256 cumulativeEquity = totalSupply
  .getDefaultPositionUnit(IERC20(component).balanceOf(address(_setToken)))
  .toInt256();
```

+ rounding up (equity) / rounding down (debt) when calculating the total units at L232

bweick:
> This looks good to me! Make sense.
> 
> To summarize, the main changes are:
> 
> * how `cumulativeEquity` is calculated on L362
> 
> ```solidity
> int256 cumulativeEquity = totalSupply
>   .getDefaultPositionUnit(IERC20(component).balanceOf(address(_setToken)))
>   .toInt256();
> ```
> 
> * rounding up (equity) / rounding down (debt) when calculating the total units at L232

Yeah to your second bullet point, it's more that in the previous implementation the same code was used for issue and redeem so we needed an `if statement` to choose which way to round, here since this piece of code is only used for issuance calculations we don't need that if statement anymore. But the rounding should still be the same for issuance vs redeem.

PR #159, opened 2021-11-03
Brian/update dependencies

bweick:
Moving packages from `devDependencies` to `dependencies` since they are used by fixtures which are used by the `set-deployments-v2` repo. Want to make sure all necessary downstream dependencies are installed in repos that consume this repo.


cgewecke (approved):
Awesome!

PR #162, opened 2021-11-12
Brian/slippage issuance

bweick:
To Do:

- [x] Add issue and redeem with new interface and checks (potentially remove `DebtIssuanceModule` dependency)
- [ ] Look at refactoring to dedupe logic
- [x] Add javadocs - particularly how to implement `getIssuanceAdjustments` and `getRedemptionAdjustments` hooks on modules
- [x] Add comments
- [x] Review tests and add any necessary scenarios for view function (adjustments that lead to negative numbers)

Potential To Dos:
- [ ] ~~Add ability to read manager hook for syncing info during view functions~~


contracts/protocol/modules/SlippageIssuanceModule.sol
```diff
@@ -0,0 +1,421 @@
...
+        internal
+        pure
+    {
+        require(_setQuantity > 0, "SetToken quantity must be > 0");
+        
+        uint256 componentsLength = _components.length;
+        require(componentsLength == _componentLimits.length, "Array length mismatch");
+        if (componentsLength > 0) {
+            require(!_components.hasDuplicate(), "Cannot duplicate addresses");
+        }
```
cgewecke, line 419: The `withSlippage`  methods don't require that checkedComponents have any elements right? Would plain issue and redemption inherited from the DIM be more substantially more efficient in that case?

I guess a case for making this method more restrictive is that someone might submit empty arrays by mistake, but can also see how multi-use is more convenient UX.

contracts/protocol/modules/SlippageIssuanceModule.sol
```diff
@@ -0,0 +1,421 @@
...
+        view
+        returns (address[] memory, uint256[] memory, uint256[] memory)
+    {
+        (
+            address[] memory components,
+            uint256[] memory equityUnits,
+            uint256[] memory debtUnits
+        ) = _getTotalIssuanceUnits(_setToken);
+
+        // NOTE: components.length isn't stored in local variable to avoid stak too deep errors. Since this function is used
```
cgewecke, line 304: stak --> stack

contracts/protocol/modules/SlippageIssuanceModule.sol
```diff
@@ -0,0 +1,421 @@
...
+            _setToken,
+            totalQuantity,
+            isIssue,
+            equityIssuanceAdjustments,
+            debtIssuanceAdjustments
+        );
+    }
+
+    /**
+     * Calculates the amount of each component  thatwill be returned on redemption net of fees as well as how much debt needs to be paid down to
```
cgewecke, line 235: thatwill --> that will

cgewecke (commented):
This is super clean and nice. 🎉 

Couple of typos in comments and left a question about param validation in #withSlippage methods but looks great to me.

cgewecke (approved):
Still a couple things getting added here but this LGTM.

contracts/protocol/modules/SlippageIssuanceModule.sol
bweick: Yeah I went back and forth on that, it wouldn't be that much more gas since it's just a couple validations then a `JUMP` opcode to call `_validateTokenTransferLimits` which wouldn't run much logic since `length=0`. This could allow a front-end integrator to only worry about one interface, though that may not be a big deal. The big con with this implementation is that it's not as explicit. May pose it to the front-end team and see what they think.

bweick: The more I think about it I think you're right, empty arrays could lead to harmful action if the front-end logic doesn't create the correct arrays. We can't stop someone from calling `issue` when they should have called `issueWithSlippage` but we can help them out of they mean to account for slippage but some logic on their end errors out.

test/protocol/modules/slippageIssuanceModule.spec.ts
```diff
@@ -0,0 +1,1658 @@
...
+      "DebtToken",
+      "DBT"
+    );
+
+    await externalPositionModule.initialize(setToken.address);
+  });
+
+  addSnapshotBeforeRestoreAfterEach();
+
+  describe("#initialize", async () => {
```
Alpha-Guy, line 82: Let's remove all un-needed tests. Tests that check parts of the code which are inherited from DIM, don't make sense here. We have already tested them. 
Let us only focus on having the tests specific to this contract, i.e. for all the overridden and new methods.
For example, the [DIMV2 tests here](https://github.com/SetProtocol/set-protocol-v2/blob/master/test/protocol/modules/debtIssuanceModuleV2.spec.ts) only focus on DIMV2 specific changes.

test/protocol/modules/slippageIssuanceModule.spec.ts
```diff
@@ -0,0 +1,1658 @@
...
+          register = true;
+        });
+
+        it("should revert", async () => {
+          await expect(subject()).to.be.revertedWith("Module not registered.");
+        });
+      });
+    });
+
+    context("External debt module has been registered with SlippageIssuanceModule", async () => {
```
Alpha-Guy, line 375: Let's only keep this and remove everything else.

contracts/protocol/modules/SlippageIssuanceModule.sol
```diff
@@ -0,0 +1,425 @@
...
+     * @param _checkedComponents    Array of components to be checked to verify required collateral doesn't exceed
+     *                                  defined max. Each entry must be unique.
+     * @param _maxTokenAmountsIn    Maps to same index in _checkedComponents. Max amount of component willing to
+     *                                  transfer in to collateralize _setQuantity amount of _setToken.
+     * @param _to                   Address to mint SetToken to
+     */
+    function issueWithSlippage(
+        ISetToken _setToken,
+        uint256 _setQuantity,
+        address[] memory _checkedComponents,
```
Alpha-Guy, line 72: I would suggest removing the `_checkedComponents` parameter. Consumers of this contract can pass in `_maxTokenAmountsIn` value as `MAX_UINT_256` for components they don't want to be checked for. It's just more intuitive, and the added simplicity outweighs the small amount of additional gas costs.

contracts/protocol/modules/SlippageIssuanceModule.sol
```diff
@@ -0,0 +1,425 @@
...
+     * to pass in a limit for every component, and may in fact only want to pass in limits for components which
+     * incur slippage to replicate (i.e. perpetuals). Passing in empty arrays for _checkComponents and
+     * _minTokenAmountsOut is equivalent to calling redeem. NOTE: not passing in limits for positions that require
+     * a trade for replication leaves the redeemer open to sandwich attacks!
+     *
+     * @param _setToken             Instance of the SetToken to redeem
+     * @param _setQuantity          Quantity of SetToken to redeem
+     * @param _checkedComponents    Array of components to be checked to verify received collateral isn't less than
+     *                                  defined min. Each entry must be unique.
+     * @param _minTokenAmountsOut   Maps to same index in _checkedComponents. Min amount of component willing to
```
Alpha-Guy, line 139: Same here. Let us remove `_minTokenAmountsOut`.

Also, once we have removed the `min/max` parameters from `issue/redeem` functions, they become exactly similar to DIM vanilla functions, which I think makes sense. Special cases falling back to vanilla cases depending on certain input parameters, is a common design pattern.

contracts/protocol/modules/SlippageIssuanceModule.sol
```diff
@@ -0,0 +1,425 @@
...
+ *
+ * The SlippageIssuanceModule is a module that enables users to issue and redeem SetTokens that requires a transaction that incurs slippage.
+ * in order to replicate the Set. Like the DebtIssuanceModule, module hooks are added to allow for syncing of positions, and component
+ * level hooks are added to ensure positions are replicated correctly. The manager can define arbitrary issuance logic in the manager hook,
+ * as well as specify issue and redeem fees. The getRequiredComponentIssuanceUnits and it's redemption counterpart now also include any
+ * changes to the position expected to happen during issuance thus providing much better estimations for positions that are synced or require
+ * a trade. It is worth noting that this module inherits from DebtIssuanceModule, consequently it can also be used for issuances that do NOT
+ * require slippage just by calling the issue and redeem endpoints.
+ */
+contract SlippageIssuanceModule is DebtIssuanceModule {
```
Alpha-Guy, line 45: Having both `issue()` and `issueWithSlippage()` functions forces the consumer to understand the differences between them, before calling tem. This might lead to confusion down the line. Are there any steps we can take to remove this confusion?

contracts/protocol/modules/SlippageIssuanceModule.sol
Alpha-Guy: Expanding on [this]() and above, does it make sense to override `redeem` to call `redeemWithSlippage()`, passing in an array filled with zeros for `_minTokenAmmountOut` argument. Same for issuance.

Alpha-Guy (commented):
There are 2 major design changes that we need to think about.
1. Where should we handle the complexity of positive/negative adjustmentent (delta) values. 
    - I suggest the DIM, instead of the module which implements the issuance hook
2. How can we make parameters passed to `issueWithSlippage() / redeemWithSlippage()` more intuitive. 
    - I suggest we drop the complexity of passing in assets to check, and default to checking all assets.
    
Also there's a lot of unwanted tests, which can be removed, and tests can focus more on the specific changes introduced in this module. I will go through the tests once again after we have cleaned it.

contracts/interfaces/IModuleIssuanceHookV2.sol
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * Adjustments should return the NET CHANGE in POSITION UNITS for each component in the SetToken's
+     * components array. Each entry in the returned arrays should index to the same component in the
+     * SetToken's components array (called using getComponents()). Directional adjustments should be made
+     * according to the following table (i.e. returning a negative debt number means debt is reduced during
+     * issue/redeem):
+     * | ------------------------------------|
+     * | Type    |  Positive |  Negative     |
+     * | -----   |---------- | --------------|
+     * | Equity  | Add Equity| Equity Reduced|
```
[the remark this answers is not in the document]
bweick: +1 agreed, great catch

test/protocol/modules/slippageIssuanceModule.spec.ts
bweick: +1

test/protocol/modules/slippageIssuanceModule.spec.ts
bweick: +1

contracts/protocol/modules/SlippageIssuanceModule.sol
bweick: Chatting with @asoong we decided we want to be explicit about the components, so I think we'd always want a components array. Given that I think it makes sense to keep as is.

contracts/protocol/modules/SlippageIssuanceModule.sol
bweick: I think we're ok with them needing to understand the difference here. They're going to have to read the docs to integrate anyway, and the interfaces are different enough it should raise questions about how each is supposed to be used

PR #164, opened 2021-11-16
Brian/slippage iss review changes

bweick:
CHANGELOG:
- Updated logic in `SlippageIssuanceModule` to expect return values from `getIssuanceAdjustments` and `getRedemptionAdjustments` to be difference in current position unit vs expected position unit after issuance
- Made small javadoc updates
- Removed some tests that were testing logic from an inherited contract

### DIFF
--- a/contracts/interfaces/IModuleIssuanceHookV2.sol
+++ b/contracts/interfaces/IModuleIssuanceHookV2.sol
@@ -1,5 +1,5 @@
 /*
-    Copyright 2020 Set Labs Inc.
+    Copyright 2021 Set Labs Inc.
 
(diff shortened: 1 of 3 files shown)


cgewecke (approved):
Looks good!

Alpha-Guy (commented):
1. Let's improve our revert string [here](https://github.com/SetProtocol/set-protocol-v2/blob/50caa3c39787ba4443fd672ca47f7aab469b9471/contracts/protocol/modules/SlippageIssuanceModule.sol#L394).
2. Any updates on how we intend to handle two issuance/redemption functions, `issue` and `issueWithSlippage` in SIM?

PR #172, opened 2021-12-01
Improve perp module tests

Alpha-Guy:
Unit Tests
- [x] Review unit tests
- [x] Add missing test cases
    - [x] #deposit - when position is long - should create a deposit
    - [x]  #moduleRedeemHook - when long one asset and short another and all it's sub-cases
    - [x] Check base token amounts when redeeming multiple Sets
    - [x]  #moduleIssueHook - # when long multiple positions - #set is owed funding
    - [x] #componentReddemHook - when short and corresponding sub-cases
- [x] Fix flaky tests
    - [x] moduleIssueHook
        - Replace looser `closeTo` checks with strict equality checks
        - Status: 24 / 24
    - [x] moduleRedeemHook
        - Replace looser `closeTo` checks with strict equality checks
        - Status: 24 / 24


utils/common/perpV2Utils.ts
```diff
@@ -103,6 +103,51 @@ export async function calculateUSDCTransferIn(
   return toUSDCDecimals(usdcAmountIn);
 }
 
+// Returns notional amount of USDC to transfer in on issue. Handles multiple positions, long and short.
+// Returned values is not convert to USDC decimals.
+export async function calculateUSDCAmountTransferIn(
```
cgewecke, line 108: What thoughts do you have about de-duplicating the logic in this file?  

We could have  `calculateUSDCTransferIn`, `calculateUSDCTransferOut ` consume this method as a base and just convert to USDC units? And maybe rename these so they convey something about the unit?

Another way would be to get rid of the USDC units version and cast everything in the test file. 

Idk...

test/protocol/modules/perpV2LeverageModule.spec.ts
```diff
@@ -1857,46 +2031,58 @@ describe("PerpV2LeverageModule", () => {
             baseToken: vETH.address,
             isBaseToQuote: false,     // long
             isExactInput: true,       // `amount` is USDC
-            amount: ether(10000),     // move price up by buying 10k USDC of vETH
+            amount: ether(1000),     // move price up by buying 1k worth of vETH
```
cgewecke, line 2034: Out of curiosity, what is the margin of error when the price shifts are more extreme?

test/protocol/modules/perpV2LeverageModule.spec.ts
```diff
@@ -1997,33 +2201,62 @@ describe("PerpV2LeverageModule", () => {
...
           // Deposit more collateral to avoid NEFC error
-          await perpLeverageModule.deposit(subjectSetToken, usdcUnits(30));
-          usdcTransferInQuantity = await calculateUSDCTransferIn(
-            setToken,
-            subjectSetQuantity,
-            perpLeverageModule,
-            perpSetup
-          );
+          // todo: confirm with chris if we don't need this any longer
+          // await perpLeverageModule.deposit(subjectSetToken, usdcUnits(30));
```
cgewecke, line 2259: Yeah, this is a relic of the first draft. Good catch.

test/protocol/modules/perpV2LeverageModule.spec.ts
```diff
@@ -3901,64 +4741,666 @@ describe("PerpV2LeverageModule", () => {
...
+
           const externalPositionUnit = await setToken.getExternalPositionRealUnit(
             usdc.address,
             perpLeverageModule.address
           );
 
-          expect(pendingFunding).lt(0);
-          expect(externalPositionUnit).to.be.closeTo(expectedExternalPositionUnit, 100);
+          // We expect pending funding to be greater than -1 USDC and discount applied
+          expect(pendingFunding).gt(0);
```
cgewecke, line 4841: Good! 👍

cgewecke (commented):
This is really nice... those perfect matches are very impressive! Just great.  

Do you think it would be possible to create helpers for the before subject / after subject time based logic and de-duplicate some of the code? 

(Possibly two helpers, one for the single asset, one for the double asset cases.)

utils/common/perpV2Utils.ts
Alpha-Guy: Fixed in [bee90b7](https://github.com/SetProtocol/set-protocol-v2/pull/172/commits/bee90b7009935f624a4a0d4cb9120cc759b11b21)

Alpha-Guy:
> This is really nice... those perfect matches are very impressive! Just great.
> 
> Do you think it would be possible to create helpers for the before subject / after subject time based logic and de-duplicate some of the code?
> 
> (Possibly two helpers, one for the single asset, one for the double asset cases.)

Added a generalized helper in [601ac](https://github.com/SetProtocol/set-protocol-v2/pull/172/commits/601ac232389a7aa55562779286fe969b569eece5)

test/protocol/modules/perpV2LeverageModule.spec.ts
Alpha-Guy: Fixed.

Alpha-Guy: But, once we set funding rate to zero, we can set the amount to any value, and the equality checks would still pass.

cgewecke (approved):
💯

PR #180, opened 2021-12-18
DG Migration Wrap V2 Adapter

FlattestWhite:
# Code Review Processes
## New Feature Review

README Checks

Code Checks
- [] Add javadocs. 

Broader Considerations


utils/deploys/deployExternal.ts
```diff
@@ -233,6 +233,9 @@ import { PerpV2InsuranceFund__factory } from "../../typechain/factories/PerpV2In
 import { PerpV2AccountBalance__factory } from "../../typechain/factories/PerpV2AccountBalance__factory";
 import { PerpV2Exchange__factory } from "../../typechain/factories/PerpV2Exchange__factory";
 
+import { Dg } from "../contracts/dg";
+import { Dg__factory } from "../../typechain/factories/Dg__factory";
```
ncitron, line 237: Are you missing the bytecode for the DG token? Looks like typechain managed to create the factory contracts so maybe you forgot to commit something? Most of the time that dg.json file would include some bytecode along with the abi.

ncitron (commented):
This is looking pretty good. One comment since I think your dg.json file is missing the bytecode. One thing we also usually do is an integration test to show that no only is the calldata generated correctly, but also that the contract interactions work as expected. Take a look at the agi migration to see how these type of tests usually work https://github.com/SetProtocol/set-protocol-v2/blob/master/test/integration/agiMigrationWrapModule.spec.ts

Also do you have the address of the new DG token?

FlattestWhite:
> This is looking pretty good. One comment since I think your dg.json file is missing the bytecode. One thing we also usually do is an integration test to show that no only is the calldata generated correctly, but also that the contract interactions work as expected. Take a look at the agi migration to see how these type of tests usually work https://github.com/SetProtocol/set-protocol-v2/blob/master/test/integration/agiMigrationWrapModule.spec.ts
> 
> Also do you have the address of the new DG token?

Ahh right - yes I'll add a test testing the before and after state.

The new DG token address is here: https://etherscan.io/token/0x4b520c812e8430659fc9f12f6d0c39026c83588d

FlattestWhite:
Screen Shot 2021-12-22 at 1 46 35 am
@ncitron need some advice - I've been trying to debug why the new integration test doens't work when it's being called through the wrap module. Do you have an idea why it's throwing this error?

ncitron:
This error usually happens when a module calls `invoke` on a set token which causes the set token to make an arbitrary call (in this case it makes the wrapping call). This call is failing and causing the error. I would start by validating that you are interacting with the DG contract correctly.

FlattestWhite:
pingaling @ncitron there's a deadline for this iirc

cgewecke:
Closing in favor of #183 (please just ping if this should remain open)

PR #182, opened 2021-12-30
Make SlippageIssuanceModule issuance & redemption component units getters non-view

cgewecke:
`getComponent...Units` methods throw when used with the PerpV2LeverageModule because its issuance and redemption adjustments getters are non-view. They simulate trades on UniswapV3 and revert in order to discover swap outcomes and can only be called off-chain via methods like `ethers.callStatic`

Have made the smallest intervention possible here, renaming the getters to reflect their intended use context and adding / updating tests. 

**Open Questions**
+ Should this PR also override the inherited getters and revert (to prevent incorrect use of those methods)
+ Or ... should we change the state-mutability signature for these methods across the inheritance chain (instead just of renaming and changing SlippageIssuance's)

### DIFF
--- a/contracts/interfaces/IModuleIssuanceHookV2.sol
+++ b/contracts/interfaces/IModuleIssuanceHookV2.sol
@@ -26,14 +26,14 @@ import { ISetToken } from "./ISetToken.sol";
  * CHANGELOG:
  *      - Added a module level issue hook that can be used to set state ahead of component level
  *        issue hooks
- *      - Added view function that return expected positional adjustments during issue/redeem to
+ *      - Added non-view getter that returns expected positional adjustments during issue/redeem to
  *        the issuance module in order to give more accurate token flow information
  */
 interface IModuleIssuanceHookV2 {
 
     function moduleIssueHook(ISetToken _setToken, uint256 _setTokenQuantity) external;
     function moduleRedeemHook(ISetToken _setToken, uint256 _setTokenQuantity) external;
-    
+
     function componentIssueHook(
         ISetToken _setToken,
         uint256 _setTokenQuantity,
@@ -53,6 +53,9 @@ interface IModuleIssuanceHookV2 {
      * components array (i.e. if debt is greater than current debt position unit return negative number).
      * Each entry in the returned arrays should index to the same component in the SetToken's components
      * array (called using getComponents()).
+     *
+     * NOTE: This getter is non-view to allow module hooks to determine units by simulating state changes in
+     * an external protocol and reverting. It should only be called by off-chain methods via static call.
      */
     function getIssuanceAdjustments(
         ISetToken _setToken,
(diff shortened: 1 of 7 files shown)


contracts/interfaces/IModuleIssuanceHookV2.sol
```diff
@@ -73,6 +72,5 @@ interface IModuleIssuanceHookV2 {
         uint256 _setTokenQuantity
     )
         external
-        view
```
richardliang, line 76: Nit: Update comment on line 29 since its no longer a view function

contracts/interfaces/IModuleIssuanceHookV2.sol
```diff
@@ -73,6 +72,5 @@ interface IModuleIssuanceHookV2 {
         uint256 _setTokenQuantity
     )
         external
```
richardliang, line 74: Are we expecting ModuleIssuanceHookV2 to be used for all modules going forward (if they require a module hook)?

contracts/protocol/modules/SlippageIssuanceModule.sol
```diff
@@ -188,23 +188,24 @@ contract SlippageIssuanceModule is DebtIssuanceModule {
...
+     *
      * @param _setToken         Instance of the SetToken to issue
      * @param _quantity         Amount of Sets to be issued
      *
      * @return address[]        Array of component addresses making up the Set
      * @return uint256[]        Array of equity notional amounts of each component, respectively, represented as uint256
      * @return uint256[]        Array of debt notional amounts of each component, respectively, represented as uint256
      */
-    function getRequiredComponentIssuanceUnits(
+    function getRequiredComponentIssuanceUnitsOffChain(
```
richardliang, line 204: Since this isn't a view function anymore, should we move this to the section above?

/* ============ External Functions ============ */

contracts/protocol/modules/SlippageIssuanceModule.sol
```diff
@@ -238,13 +242,11 @@ contract SlippageIssuanceModule is DebtIssuanceModule {
      * @return uint256[]        Array of equity notional amounts of each component, respectively, represented as uint256
      * @return uint256[]        Array of debt notional amounts of each component, respectively, represented as uint256
      */
-    function getRequiredComponentRedemptionUnits(
+    function getRequiredComponentRedemptionUnitsOffChain(
```
richardliang, line 245: Since this isn't a view function anymore, should we move this to the section above?

/* ============ External Functions ============ */

richardliang (approved):
LGTM, just had a question on the module hook v2

contracts/protocol/modules/SlippageIssuanceModule.sol
cgewecke: 431c3e0

contracts/protocol/modules/SlippageIssuanceModule.sol
cgewecke: 431c3e0

contracts/interfaces/IModuleIssuanceHookV2.sol
cgewecke: 0bbccbc

richardliang (approved):

PR #183, opened 2021-12-31
DG Migration Adapter

ncitron:
This builds off of @flattestwhite's #180 except it uses the original `WrapModule` instead of `WrapModuleV2`

DG is doing a token migration with a 1 to 1000 token split. The address of the new "DGLight" token is [0x4b520c812e8430659fc9f12f6d0c39026c83588d](https://etherscan.io/token/0x4b520c812e8430659fc9f12f6d0c39026c83588d). The new token also acts as the migration contract.


test/protocol/integration/wrap/dgMigrationWrapAdapter.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+        subjectUnderlyingAddress,
+        subjectWrappedAddress
+      );
+    }
+    it("should have the correct legacy token address", async () => {
+      const deployedAdapter = await subject();
+
+      expect(await deployedAdapter.dgTokenV1()).to.eq(subjectUnderlyingAddress);
+    });
+    it("should have the correct new token address", async () => {
```
cgewecke, line 68: Nit: space

cgewecke (approved):
This LGTM! Seems straightforward. 

(One thing that would be nice to have here are some links in the PR description to info and context about this migration. Like ... what's the difference between these two tokens etc?)

ncitron:
Fixed the spacing issue and added a bit of context in the PR description.

PR #187, opened 2022-01-12
Retain hardcoded gas settings in published artifacts

cgewecke:
## DO NOT MERGE

This PR reverts a publication policy we had of stripping out hardcoded gas values in the published artifacts / ABIs. 
In set-v2-strategies we are consuming this repo's typechain folder directly from node_modules and are unable to modify those artifacts. The extremely slow execution speeds are making the tests unstable.

While it's not ideal to ship with these hardcoded gas settings, other Set contract repos with the same repo design as this one are their principal consumer *as a package* (and they need this) 

This change should not affect:
+ the set-v2-deployment repo since it explicitly copies and recompiles the contracts into the project root on build. 
+ the set-ui repo since they use abi fragments and manage their own gas estimation


cgewecke:
Actually ... this isn't safe for set.js since that's used by external users. We have to find another solution to this problem.

cgewecke:
Closing. We'll continue to use this branch to publish set-v2 in the format required to get set-v2-strategies tests to pass but we can't merge these changes because it's unsafe to publish ABIs with hardcoded gas values.

People relying on ethers gas estimation to catch failing transactions could instead begin automatically sending bad txs and burn $$$.

PR #190, opened 2022-01-24
PerpV2LeverageModule ABDK audit fixes

Alpha-Guy:
https://docs.google.com/spreadsheets/d/1-S4Gkn5UD5MsZcODgCHCOBY9UX8bETPEk7jzzvwxsYY/edit#gid=0

### DIFF
--- a/contracts/interfaces/IPerpV2LeverageModule.sol
+++ b/contracts/interfaces/IPerpV2LeverageModule.sol
@@ -75,7 +75,7 @@ interface IPerpV2LeverageModule {
      * @param _protocolFee      Quantity in collateral decimals sent to fee recipient during lever trade
      * @param _isBuy            True when baseToken is being bought, false when being sold
      */
-    event PerpTraded(
+    event PerpTrade(
         ISetToken indexed _setToken,
         address indexed _baseToken,
         uint256 _deltaBase,
@@ -90,7 +90,7 @@ interface IPerpV2LeverageModule {
      * @param _collateralToken      Token being deposited as collateral (USDC)
      * @param _amountDeposited      Amount of collateral being deposited into Perp
      */
-    event CollateralDeposited(
+    event CollateralDeposit(
         ISetToken indexed _setToken,
         IERC20 _collateralToken,
         uint256 _amountDeposited
@@ -102,7 +102,7 @@ interface IPerpV2LeverageModule {
      * @param _collateralToken      Token being withdrawn as collateral (USDC)
      * @param _amountWithdrawn      Amount of collateral being withdrawn from Perp
      */
-    event CollateralWithdrawn(
+    event CollateralWithdrawl(
         ISetToken indexed _setToken,
         IERC20 _collateralToken,
         uint256 _amountWithdrawn
--- a/contracts/interfaces/IPerpV2LeverageModule.sol
+++ b/contracts/interfaces/IPerpV2LeverageModule.sol
@@ -108,6 +108,32 @@ interface IPerpV2LeverageModule {
         uint256 _amountWithdrawn
     );
 
+    /* ============ State Variable Getters ============ */
+
+    // PerpV2 contract which provides getters for base, quote, and owedRealizedPnl balances
+    function perpAccountBalance() external view returns(IAccountBalance);
+
+    // PerpV2 contract which provides a trading API
+    function perpClearingHouse() external view returns(IClearingHouse);
+
+    // PerpV2 contract which manages trading logic. Provides getters for UniswapV3 pools and pending funding balances
+    function perpExchange() external view returns(IExchange);
+
+    // PerpV2 contract which handles deposits and withdrawals. Provides getter for collateral balances
+    function perpVault() external view returns(IVault);
+
+    // PerpV2 contract which makes it possible to simulate a trade before it occurs
+    function perpQuoter() external view returns(IQuoter);
+
+    // PerpV2 contract which provides a getter for baseToken UniswapV3 pools
+    function perpMarketRegistry() external view returns(IMarketRegistry);
+
+    // Token (USDC) used as a vault deposit, Perp currently only supports USDC as it's settlement and collateral token
+    function collateralToken() external view returns(IERC20);
+
+    // Decimals of collateral token. We set this in the constructor for later reading
+    function collateralDecimals() external view returns(uint8);
+    
     /* ============ External Functions ============ */
 
     /**
@@ -263,11 +289,4 @@ interface IPerpV2LeverageModule {
      * @return price                Mid-point price of virtual token in UniswapV3 AMM market
      */
     function getAMMSpotPrice(address _baseToken) external view returns (uint256 price);
-
-    /**
-     * @dev Returns address of collateral token used by Perpetual Protocol (USDC)
-     *
-     * @return                      Address of Perpetual Protocol collateral token (USDC)
-     */
-    function collateralToken() external view returns (IERC20);
 }
--- a/contracts/interfaces/IPerpV2LeverageModule.sol
+++ b/contracts/interfaces/IPerpV2LeverageModule.sol
@@ -75,7 +75,7 @@ interface IPerpV2LeverageModule {
      * @param _protocolFee      Quantity in collateral decimals sent to fee recipient during lever trade
      * @param _isBuy            True when baseToken is being bought, false when being sold
      */
-    event PerpTrade(
+    event PerpTraded(
         ISetToken indexed _setToken,
         address indexed _baseToken,
         uint256 _deltaBase,
@@ -90,7 +90,7 @@ interface IPerpV2LeverageModule {
      * @param _collateralToken      Token being deposited as collateral (USDC)
      * @param _amountDeposited      Amount of collateral being deposited into Perp
      */
-    event CollateralDeposit(
+    event CollateralDeposited(
         ISetToken indexed _setToken,
         IERC20 _collateralToken,
         uint256 _amountDeposited
@@ -102,7 +102,7 @@ interface IPerpV2LeverageModule {
      * @param _collateralToken      Token being withdrawn as collateral (USDC)
      * @param _amountWithdrawn      Amount of collateral being withdrawn from Perp
      */
-    event CollateralWithdrawl(
+    event CollateralWithdrawn(
         ISetToken indexed _setToken,
         IERC20 _collateralToken,
         uint256 _amountWithdrawn
--- a/contracts/protocol/modules/PerpV2LeverageModule.sol
+++ b/contracts/protocol/modules/PerpV2LeverageModule.sol
@@ -249,7 +249,13 @@ contract PerpV2LeverageModule is ModuleBase, ReentrancyGuard, Ownable, SetTokenA
         // Try if register exists on any of the modules including the debt issuance module
         address[] memory modules = _setToken.getModules();
         for(uint256 i = 0; i < modules.length; i++) {
-            try IDebtIssuanceModule(modules[i]).registerToIssuanceModule(_setToken) {} catch {}
+            try IDebtIssuanceModule(modules[i]).registerToIssuanceModule(_setToken) {
+                // This module registered itself on `modules[i]` issuance module.
+            } catch {
+                // Try will fail if `modules[i]` is not an instance of IDebtIssuanceModule and does not
+                // implement the `registerToIssuanceModule` function, or if the `registerToIssuanceModule`
+                // function call reverted. Irrespective of the reason for failure, continue to the next module.
+            }
         }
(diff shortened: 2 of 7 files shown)


contracts/interfaces/IPerpV2LeverageModule.sol
```diff
@@ -75,7 +75,7 @@ interface IPerpV2LeverageModule {
      * @param _protocolFee      Quantity in collateral decimals sent to fee recipient during lever trade
      * @param _isBuy            True when baseToken is being bought, false when being sold
      */
-    event PerpTraded(
+    event PerpTrade(
```
bweick, line 78: Let's revert these naming changes. This is one of those things where we've deviated stylistically from ABDK. We use verbs for the action instead of nouns.

contracts/protocol/modules/PerpV2LeverageModule.sol
```diff
@@ -142,10 +142,10 @@ contract PerpV2LeverageModule is ModuleBase, ReentrancyGuard, Ownable, SetTokenA
      * @param _collateralToken      Token being withdrawn as collateral (USDC)
      * @param _amountWithdrawn      Amount of collateral being withdrawn from Perp
      */
-    event CollateralWithdrawn(
+    event CollateralWithdrawl(
```
bweick, line 145: Same here as well as above

contracts/protocol/modules/PerpV2LeverageModule.sol
```diff
@@ -599,10 +602,11 @@ contract PerpV2LeverageModule is ModuleBase, ReentrancyGuard, Ownable, SetTokenA
      *         + quoteBalance: USDC quote asset balance as notional quantity (10**18)
      */
     function getPositionNotionalInfo(ISetToken _setToken) public view returns (PositionNotionalInfo[] memory) {
-        PositionNotionalInfo[] memory positionInfo = new PositionNotionalInfo[](positions[_setToken].length);
+        address[] memory positionList = positions[_setToken];
```
bweick, line 605: I think here' we're just better off storing `positionInfo.length` into it's own variable `positionLength`. Then we can reuse `positionLength` throughout the function. There's no need to load `positionList` into memory since that's used to build the `positionInfo` array.

contracts/protocol/modules/PerpV2LeverageModule.sol
```diff
@@ -749,10 +730,11 @@ contract PerpV2LeverageModule is ModuleBase, ReentrancyGuard, Ownable, SetTokenA
         // and variable may refer to the value which will be redeemed.
         int256 accountValueIssued = _calculatePartialAccountValuePositionUnit(_setToken).preciseMul(setTokenQuantityInt);
 
-        PositionUnitInfo[] memory positionInfo = getPositionUnitInfo(_setToken);
+        PositionNotionalInfo[] memory positionInfo = getPositionNotionalInfo(_setToken);
```
bweick, line 733: Why was this edit made?

contracts/protocol/modules/PerpV2LeverageModule.sol
```diff
@@ -1100,8 +1084,9 @@ contract PerpV2LeverageModule is ModuleBase, ReentrancyGuard, Ownable, SetTokenA
         address[] memory positionList = positions[_setToken];
 
         for (uint256 i = 0; i < positionList.length; i++) {
-            if (!_hasBaseBalance(_setToken, positionList[i])) {
-                positions[_setToken].removeStorage(positionList[i]);
+            address currPosition = positionList[i];
```
bweick, line 1087: Let's name this a bit better, `currentBaseToken`. The current name made me think we were removing at a certain index which would be a problem in this function.

contracts/interfaces/IPerpV2LeverageModule.sol
Alpha-Guy: [95db13](https://github.com/SetProtocol/set-protocol-v2/pull/190/commits/95db136944e08ff2b0598da2bf6bf516e7dc3b59)

contracts/protocol/modules/PerpV2LeverageModule.sol
Alpha-Guy: [95db13](https://github.com/SetProtocol/set-protocol-v2/pull/190/commits/95db136944e08ff2b0598da2bf6bf516e7dc3b59)

contracts/protocol/modules/PerpV2LeverageModule.sol
Alpha-Guy: Yeah we can store length in a separate variable. But by removing the `positionList`, we will have `address baseToken = positions[_setToken][i]`, and the `positions[_setToken]` would be fetched during each iteration of the for loop.

contracts/protocol/modules/PerpV2LeverageModule.sol
Alpha-Guy: Because, I modified the `getPositionUnitInfo` function to use the `getPositionNotionalInfo` function. Now, we are iterating through the `positions` array 2 times in total by the `getPositionNotionalInfo` and calculating unit in the below for loop. Calling `getPositionUnitInfo` will lead to 3 iterations.

contracts/protocol/modules/PerpV2LeverageModule.sol
Alpha-Guy: But it's not `currBaseToken`, it's a position. However I update the name to `currentPosition` for better clarity in [3f283aa](https://github.com/SetProtocol/set-protocol-v2/pull/190/commits/3f283aa4204c25a8e3dbf77db4770ccee728b5b9).

contracts/protocol/modules/PerpV2LeverageModule.sol
Alpha-Guy: Updated all for loops to calculate the number of iterattions (length) before and not in the for loop condition in [a8cffd4](https://github.com/SetProtocol/set-protocol-v2/pull/190/commits/a8cffd41d6bdfcf542b0ad993fb5c6d6baeb3fa3). It is more gas efficient.

cgewecke (commented):
Questions about skipped items from the ABDK audit:

**Item 18, L757**
In two places we're using `.mul(-1)` to flip the sign. ABDK says it's more optimal to use a safe negation function. 

+ https://github.com/SetProtocol/set-protocol-v2/blob/4892ef8a76ef41b36059f54d96a8b0126be08453/contracts/protocol/modules/PerpV2LeverageModule.sol#L678
+ https://github.com/SetProtocol/set-protocol-v2/blob/4892ef8a76ef41b36059f54d96a8b0126be08453/contracts/protocol/modules/PerpV2LeverageModule.sol#L762

At Perp, negation is handled using a set of methods that look like this:
```solidity
function neg256(int256 a) internal pure returns (int256) {
        require(a > -2**255, "PerpMath: inversion overflow");
        return -a;
    }

    function neg256(uint256 a) internal pure returns (int256) {
        return -PerpSafeCast.toInt256(a);
    }

    function neg128(int128 a) internal pure returns (int128) {
        require(a > -2**127, "PerpMath: inversion overflow");
        return -a;
    }

    function neg128(uint128 a) internal pure returns (int128) {
        return -PerpSafeCast.toInt128(a);
    }
 ```

+ [PerpMath.sol][1]
+ [PerpSafeCast][2]

[1]: https://github.com/perpetual-protocol/perp-lushan/blob/744d16bd71439ce2b3bb9dc730f9f0e3ae3a5da2/contracts/lib/PerpMath.sol#L40
[2]: https://github.com/perpetual-protocol/perp-lushan/blob/744d16bd71439ce2b3bb9dc730f9f0e3ae3a5da2/contracts/lib/PerpSafeCast.sol#L8

**Question**: We are using SafeMath for `mul` so there shouldn't be any overflow issues but what would the efficiency gain proposed in this recommendation look like in our case and what is our reason for skipping? (Fine w/ skipping fwiw - this flip is simple). 

------
**Item 11: try/catch blocks**

ABDK says:
> Consider ignoring only a certain class of errors or implementing some way to detect which modules do implement the “IDebtIssuanceModule” interface.

Their suggestion might not be viable but could we add a clearer comment about what's happening here, e.g why it's wrapped like this, etc? 

https://github.com/SetProtocol/set-protocol-v2/blob/4892ef8a76ef41b36059f54d96a8b0126be08453/contracts/protocol/modules/PerpV2LeverageModule.sol#L249-L254

File: contracts/protocol/modules/PerpV2LeverageModule.sol (hunk 3 above)
Comment (bweick) [reply contracts/protocol/modules/PerpV2LeverageModule.sol:605]: Ok here we should store the `positionList.length` as a variable and then use it to instantiate the `positionInfo` array and set as the bound on the `for...loop`. You're right about needing to load `positionList` into memory, I misread.

File: contracts/protocol/modules/PerpV2LeverageModule.sol (hunk 4 above)
Comment (bweick) [reply contracts/protocol/modules/PerpV2LeverageModule.sol:735]: ok. I guess in that case we aren't using `getPositionUnitInfo` internally anymore. May make sense to make that function have `external` visibility.

File: contracts/protocol/modules/PerpV2LeverageModule.sol [hunk 6]
```diff
@@ -1098,10 +1085,12 @@ contract PerpV2LeverageModule is ModuleBase, ReentrancyGuard, Ownable, SetTokenA
      */
     function _syncPositionList(ISetToken _setToken) internal {
         address[] memory positionList = positions[_setToken];
-
-        for (uint256 i = 0; i < positionList.length; i++) {
-            if (!_hasBaseBalance(_setToken, positionList[i])) {
-                positions[_setToken].removeStorage(positionList[i]);
+        uint positionLength = positionList.length;
```
bweick, line 1088: let's make this `uint256` for consistency across the codebase

contracts/protocol/modules/PerpV2LeverageModule.sol
Alpha-Guy: Done in [d77566](https://github.com/SetProtocol/set-protocol-v2/pull/190/commits/d7756688ae432d06ed69d66976e9d0c9dbc15608).

contracts/protocol/modules/PerpV2LeverageModule.sol
Alpha-Guy: Done in [c04b986](https://github.com/SetProtocol/set-protocol-v2/pull/190/commits/c04b986dec97c6414f3802cff0752786fbb5ee8f)

Alpha-Guy:
@cgewecke  `neg()` takes 499 and `mul()` takes 624.
More gas used in `mul(int256 a, int256 b)` can be due to bunch of checks for both `a`, `b` and `c` (variable used to store result).
Updated to using `neg()` in [d8e362f](https://github.com/SetProtocol/set-protocol-v2/pull/190/commits/d8e362fff2e1732e421eace8c0b57f4ebedad0f2).
Refactored `neg()` and added tests in [4d7f6e2](https://github.com/SetProtocol/set-protocol-v2/pull/190/commits/4d7f6e2784c07cc89a76829ddcfcfa2427a709b5)

And added comments for the try-catch block in [0e704f03](https://github.com/SetProtocol/set-protocol-v2/pull/190/commits/0e704f03731e824a4164fc76cacbe76b63debabe)

Alpha-Guy:
> Questions about skipped items from the ABDK audit:
> **Item 18, L757**
> **Question**: We are using SafeMath for `mul` so there shouldn't be any overflow issues but what would the efficiency gain proposed in this recommendation look like in our case and what is our reason for skipping? (Fine w/ skipping fwiw - this flip is simple).
> 
> **Item 11: try/catch blocks**
> Their suggestion might not be viable but could we add a clearer comment about what's happening here, e.g why it's wrapped like this, etc?
> 
> https://github.com/SetProtocol/set-protocol-v2/blob/4892ef8a76ef41b36059f54d96a8b0126be08453/contracts/protocol/modules/PerpV2LeverageModule.sol#L249-L254

@cgewecke  `neg()` takes 499 and `mul()` takes 624 gas.
More gas used in `mul(int256 a, int256 b)` can be attributed to a bunch of checks for both `a`, `b` and `c` (variable used to store result).
Updated to using `neg()` in [d8e362f](https://github.com/SetProtocol/set-protocol-v2/pull/190/commits/d8e362fff2e1732e421eace8c0b57f4ebedad0f2).
Refactored `neg()` and added tests in [4d7f6e2](https://github.com/SetProtocol/set-protocol-v2/pull/190/commits/4d7f6e2784c07cc89a76829ddcfcfa2427a709b5)

And added comments for the try-catch block in [0e704f03](https://github.com/SetProtocol/set-protocol-v2/pull/190/commits/0e704f03731e824a4164fc76cacbe76b63debabe)

PR #191, opened 2022-01-30
fix bug in function preciseDivCeil

gzliudan:
There's a bug in current version:
```solidity
    function preciseDivCeil(int256 a, int256 b) internal pure returns (int256) {
        require(b != 0, "Cant divide by 0");
        
        a = a.mul(PRECISE_UNIT_INT);
        int256 c = a.div(b);
        if (c * b != a) {
            c = c > 0 ? c + 1 : c - 1;
        }
        return c;
    }
```
For example: when a is 1, b is 1000000000000000000000, preciseDivCei should return 1. But in function preciseDivCeil, the value  of  c = a.div(b) is 0, and since c * b != a is true, c > 0 is false, then c = c - 1 = -1, which is not right!

### DIFF
--- a/contracts/lib/PreciseUnitMath.sol
+++ b/contracts/lib/PreciseUnitMath.sol
@@ -147,9 +147,11 @@ library PreciseUnitMath {
         
         a = a.mul(PRECISE_UNIT_INT);
         int256 c = a.div(b);
-        if (c * b != a) {
-            c = c > 0 ? c + 1 : c - 1;
+
+        if (a % b != 0) {
+            (a ^ b > 0) ? ++c : --c;
         }
+
         return c;
     }
 
--- a/utils/common/mathUtils.ts
+++ b/utils/common/mathUtils.ts
@@ -39,14 +39,15 @@ export const preciseDivCeil = (a: BigNumber, b: BigNumber): BigNumber => {
 };
 
 export const preciseDivCeilInt = (a: BigNumber, b: BigNumber): BigNumber => {
-  if (a.eq(0) || b.eq(0)) {
-    return ZERO;
+  const result = a.mul(PRECISE_UNIT).div(b);
+  if (result.mul(b).eq(a.mul(PRECISE_UNIT))) {
+    return result;
   }
 
-  if (a.gt(0) && b.gt(0) || a.lt(0) && b.lt(0)) {
-    return a.mul(PRECISE_UNIT).sub(1).div(b).add(1);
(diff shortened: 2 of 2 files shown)


gzliudan:
test function preciseDivCeilInt in file utils/common/mathUtils.ts returns wrong result sometimes, such as below test cases:
```js
preciseDivCeil(16000000000000000000, -1)
preciseDivCeil(15000000000000000000, -1000)
preciseDivCeil(14000000000000000000, -1000000)
preciseDivCeil(13000000000000000000, -1000000000)
preciseDivCeil(12000000000000000000, -1000000000000)
preciseDivCeil(11000000000000000000, -1000000000000000)
preciseDivCeil(10000000000000000000, -1000000000000000000)
preciseDivCeil(8000000000000000000, -1000000000000000000000)
preciseDivCeil(7000000000000000000, -1000000000000000000000000)
preciseDivCeil(6000000000000000000, -1000000000000000000000000000)
preciseDivCeil(5000000000000000000, -1000000000000000000000000000000)
preciseDivCeil(4000000000000000000, -1000000000000000000000000000000000)
preciseDivCeil(3000000000000000000, -1000000000000000000000000000000000000)
```

contracts/lib/PreciseUnitMath.sol
```diff
@@ -147,9 +147,11 @@ library PreciseUnitMath {
         
         a = a.mul(PRECISE_UNIT_INT);
         int256 c = a.div(b);
-        if (c * b != a) {
-            c = c > 0 ? c + 1 : c - 1;
+
+        if (a % b != 0) {
+            (a ^ b > 0) ? ++c : --c;
```
bweick, line 152: Great catch! One small nit here, can we add a quick comment that that the `a ^ b == 0` case is covered by the previous `if` statement so it won't resolve to `--c`. Want to make sure that context is preserved.

bweick:
@gzliudan thanks for finding this. Would definitely be interested in talking with you further if you're open to it. Feel free to e-mail me at brian@setprotocol.com.

contracts/lib/PreciseUnitMath.sol
gzliudan: Yes, It's better to add some comments.

bweick (approved):

bweick:
@gzliudan I merged this because I want to get it in a round of updates we're pushing soon. We'll add the comment I mentioned above. Look forward to chatting more with you!

PR #192, opened 2022-02-01
PerpV2 module fixes and optimizations

0xSachinK:


contracts/protocol/modules/PerpV2LeverageModule.sol
```diff
@@ -1078,6 +1099,7 @@ contract PerpV2LeverageModule is ModuleBase, ReentrancyGuard, Ownable, SetTokenA
                 positions[_setToken].removeStorage(_baseToken);
             }
         } else {
+            require(positions[_setToken].length < maxPerpPositionsPerSet, "Exceeds max perpetual positions per set");
```
bweick, line 1102: Ideally we could check this before executing the trade (since this is only called in trade). Is there a way that we can determine the `hasBaseToken` variable earlier for a validation and then pass it down to the `_updatePositionList` function?

test/protocol/modules/perpV2LeverageModule.spec.ts
```diff
@@ -862,7 +900,40 @@ describe("PerpV2LeverageModule", () => {
             });
           });
 
-          // todo: a test case which tries to flip the position?
```
bweick, line 865: I think we need a test for closing the position. We should check to make sure that when the exact negation of a baseToken position unit is passed in that it will completely zero out our vAsset position

contracts/protocol/modules/PerpV2LeverageModule.sol
```diff
@@ -183,6 +183,9 @@ contract PerpV2LeverageModule is ModuleBase, ReentrancyGuard, Ownable, SetTokenA
     // PerpV2 contract which provides a getter for baseToken UniswapV3 pools
     IMarketRegistry public immutable perpMarketRegistry;
 
+    // Max perpetual positions per SetToken
```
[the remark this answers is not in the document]
0xSachinK: Done in [dc4dd8c](https://github.com/SetProtocol/set-protocol-v2/pull/192/commits/dc4dd8c496e686435a0409c60a6c8a59827b3907)

contracts/protocol/modules/PerpV2LeverageModule.sol
```diff
@@ -299,13 +322,23 @@ contract PerpV2LeverageModule is ModuleBase, ReentrancyGuard, Ownable, SetTokenA
         nonReentrant
         onlyManagerAndValidSet(_setToken)
     {
-        ActionInfo memory actionInfo = _createAndValidateActionInfo(
+        _validateTrade(_baseToken, _baseQuantityUnits);
+
+        int256 baseBalance = perpAccountBalance.getBase(address(_setToken), _baseToken);
```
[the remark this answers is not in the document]
0xSachinK: Done in [c1b14e7](https://github.com/SetProtocol/set-protocol-v2/pull/192/commits/c1b14e705d9c077292eec318873bab50065e4668). Basically brought back the `createAndValidateActionInfo` function and added those lines to that function.

test/protocol/modules/perpV2LeverageModule.spec.ts
0xSachinK: Tests for closing a position already exist. Added checks to ensure the base balance is zero to those tests in [4efcab](https://github.com/SetProtocol/set-protocol-v2/pull/192/commits/4efcabf20fdbd7482d350812ca851c23eb21e3e0)

contracts/protocol/modules/PerpV2LeverageModule.sol
0xSachinK: That would decrease "separation of concerns" between the external and internal function. Also, it is on L2, where computation is cheap, so it shouldn't be a major problem.

bweick (approved):

PR #198, opened 2022-03-02
PerpV2 Basis Trading Module

0xSachinK:


contracts/protocol/modules/PerpV2BasisTradingModule.sol
```diff
@@ -0,0 +1,507 @@
...
+    function withdrawFundingAndAccrueFees(
+        ISetToken _setToken,
+        uint256 _amount,
+        bool _trackSettledFunding
+    )
+        external
+        nonReentrant
+        onlyManagerAndValidSet(_setToken)   
+    {
+        _updateSettledFunding(_setToken, _trackSettledFunding);
```
richardliang, line 209: If `_trackSettledFunding` is true, it would be very difficult to calculate the correct `_amount` to withdraw all settled funding right.

Maybe can create some better UX i.e. if _amount > settledFunding then withdraw settledFunding

contracts/protocol/modules/PerpV2BasisTradingModule.sol
```diff
@@ -0,0 +1,507 @@
...
+        override
+        returns (int256[] memory, int256[] memory _)
+    {
+        address[] memory components = _setToken.getComponents();
+
+        if (positions[_setToken].length > 0) {
+            int256 newExternalPositionUnit = _executePositionTrades(_setToken, _setTokenQuantity, false, true);
+
+            // Calculate performance fee unit
+            uint256 performanceFeeUnit = _getUpdateSettledFunding(_setToken)
```
richardliang, line 389: Add more comments here to explain the calculations

contracts/protocol/modules/PerpV2BasisTradingModule.sol
```diff
@@ -0,0 +1,507 @@
...
+     * @dev Updates tracked settled funding. Once funding is settled to `owedRealizedPnl` on Perpetual protocol, it is difficult to 
+     * extract out the funding value again on-chain. This function is called in an external function and is used to track and store 
+     * pending funding payment that is about to be settled due to subsequent logic in the external function.
+     *
+     * @param _setToken             Instance of SetToken
+     * @param _trackSettledFunding  Updates tracked settled funding if true
+     */
+    function _updateSettledFunding(ISetToken _setToken, bool _trackSettledFunding) internal {
+        if (_trackSettledFunding) {
+            settledFunding[_setToken] = _getUpdateSettledFunding(_setToken);
```
richardliang, line 414: Can inline this if statement

contracts/protocol/modules/PerpV2BasisTradingModule.sol
```diff
@@ -0,0 +1,507 @@
...
+        emit FeeRecipientUpdated(_setToken, _newFeeRecipient);
+    }
+
+
+    /* ============ External Getter Functions ============ */
+
+    /**
+     * @dev Gets the positive equity collateral externalPositionUnit that would be calculated for
+     * redeeming a quantity of SetToken representing the amount of collateral returned per SetToken.
+     * Values in the returned arrays map to the same index in the SetToken's components array.
```
richardliang, line 368: Add a bit more context where/why this function is used

test/protocol/modules/perpV2BasisTradingModule.spec.ts
```diff
@@ -0,0 +1,1673 @@
...
+        it("should enable the Module on the SetToken", async () => {
+          await subject();
+          const isModuleEnabled = await setToken.isInitializedModule(perpBasisTradingModule.address);
+          expect(isModuleEnabled).to.eq(true);
+        });
+      });
+    });
+  });
+
+  describe("#trade", () => {
```
richardliang, line 469: `tradeAndTrackFunding`

richardliang (changes requested):
As we discussed in DMs
* Performance fees should be a fee split vs a flat fee
* We can remove the `_trackSettledFunding` param in the `tradeAndTrackFunding` function

Only looked at the changelog in the issuance hooks and getRedemptionAdjustments. Will take another pass after these changes

contracts/protocol/modules/PerpV2BasisTradingModule.sol
```diff
@@ -0,0 +1,339 @@
...
+// 2. Add javadocs.
+
+/**
+ * @title PerpV2BasisTradingModule
+ * @author Set Protocol
+ *
+ * NOTE: The external position unit is only updated on an as-needed basis during issuance/redemption. It does not reflect the current
+ * value of the Set's perpetual position. The current value can be calculated from getPositionNotionalInfo.
+ */
+contract PerpV2BasisTradingModule is PerpV2LeverageModule {
```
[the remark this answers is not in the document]
0xSachinK: The naming is geared towards the product category it enables and it is easier to use this name across the rest of the tech stack. We had a very similar discussion last time, where we wrote DIMV2 and named it DIMV2, although other parts of the stack said it would have been easier had it just been called AaveIssuanceModule.
Also, PerpV2LeverageModuleV2 could be used for the module which supports multicollateral, and if we name this V2, we would have to name the next one V3, and it becomes tougher to manage down the line.

contracts/protocol/modules/PerpV2BasisTradingModule.sol
0xSachinK: Good suggestion.

contracts/lib/UnitConversionUtils.sol
```diff
@@ -67,4 +67,15 @@ library UnitConversionUtils {
...
     }
+
+    /**
+     * @dev Converts an arbitrarily decimalized quantity into a PRECISE_UNIT quantity.
+     *
+     * @param _amount       Non-PRECISE_UNIT amount to convert
+     * @param _decimals     Decimal precision of amount being converted to PRECISE_UNIT
+     * @return              Input converted to PRECISE_UNIT decimal format
+     */
+    function toPreciseUnitsFromDecimals(uint256 _amount, uint8 _decimals) internal pure returns (uint256) {
```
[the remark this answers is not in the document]
0xSachinK: It is consistent with the other two function names, `fromPreciseUnitToDecimals`.

contracts/protocol/modules/PerpV2BasisTradingModule.sol
```diff
@@ -0,0 +1,339 @@
...
+    function _handleFees(
+        ISetToken _setToken,
+        uint256 _amount
+    )
+        internal
+        returns (uint256 managerTake, uint256 protocolTake, uint256 totalFees)
+    {
+        uint256 performanceFee = feeStates[_setToken].performanceFeePercentage;
+
+        if (performanceFee > 0) {
```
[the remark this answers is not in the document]
bweick: Feels like the fee should be a percentage of the fees accrued by the manager imo

contracts/protocol/modules/PerpV2BasisTradingModule.sol
```diff
@@ -0,0 +1,520 @@
...
+    }
+
+    /**
+     * @dev Validates fee settings.
+     *
+     * @param _settings     FeeState struct containing performance fee settings
+     */
+    function _validateFeeState(FeeState memory _settings) internal view {
+        require(_settings.feeRecipient != address(0), "Fee Recipient must be non-zero address");
+        require(_settings.maxPerformanceFeePercentage < PreciseUnitMath.preciseUnit(), "Max fee must be < 100%");
```
bweick, line 498: <=

contracts/protocol/modules/PerpV2BasisTradingModule.sol
```diff
@@ -0,0 +1,520 @@
...
+    /**
+     * @dev Adds pending funding payment to tracked settled funding. Returns updated settled funding value.
+     * 
+     * NOTE: Tracked settled funding value can not be less than zero, hence it is reset to zero if pending funding
+     * payment is negative and |pending funding payment| >= |settledFunding[_setToken]|.
+     *
+     * @param _setToken             Instance of SetToken     
+     */
+    function _getUpdateSettledFunding(ISetToken _setToken) internal returns (uint256) {
+        // todo: Following call fetches pending funding payment across all markets. Should we change it?
```
bweick, line 440: What are the alternatives?

contracts/protocol/modules/PerpV2BasisTradingModule.sol
```diff
@@ -0,0 +1,520 @@
...
+    function updatePerformanceFee(
+        ISetToken _setToken,
+        uint256 _newFee
+    )
+        external
+        onlyManagerAndValidSet(_setToken)
+    {
+        require(_newFee < _maxPerformanceFeePercentage(_setToken), "Fee must be less than max");
+        
+        // todo: call updateSettledFunding here?
```
bweick, line 352: should definitely accrue fees here

contracts/protocol/modules/PerpV2BasisTradingModule.sol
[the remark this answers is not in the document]
0xSachinK: Fixed in [8376fb92](https://github.com/SetProtocol/set-protocol-v2/pull/198/commits/8376fb9210df683d5fa25115b61728aea3417e06).

contracts/protocol/modules/PerpV2BasisTradingModule.sol
0xSachinK: Added in [db0dd4](https://github.com/SetProtocol/set-protocol-v2/pull/198/commits/db0dd4ec2a1c36fdab64eb3172c909d28495287c)

contracts/protocol/modules/PerpV2BasisTradingModule.sol
```diff
@@ -0,0 +1,520 @@
...
+        // e.g a positive number is a debt which gets subtracted from owedRealizedPnl on settlement.
+        // We are flipping its sign here to reflect its settlement value.
+        int256 pendingFundingToBeSettled =  perpExchange.getAllPendingFundingPayment(address(_setToken)).neg();
+        
+        if (pendingFundingToBeSettled > 0) {
+            return settledFunding[_setToken].add(pendingFundingToBeSettled.toUint256());
+        }
+
+        if (settledFunding[_setToken] > pendingFundingToBeSettled.abs()) {
+            return settledFunding[_setToken].sub(pendingFundingToBeSettled.abs());
```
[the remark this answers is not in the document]
0xSachinK: Yeah the return zero at the end handles that case. (Assuming `settledFunding[_setToken]` doesn't go below zero.)

contracts/protocol/modules/PerpV2BasisTradingModule.sol
```diff
@@ -0,0 +1,511 @@
...
+    function withdrawFundingAndAccrueFees(
+        ISetToken _setToken,
+        uint256 _notionalFunding,
+        bool _trackSettledFunding
+    )
+        external
+        nonReentrant
+        onlyManagerAndValidSet(_setToken)
+    {
+        _updateSettledFunding(_setToken, _trackSettledFunding);
```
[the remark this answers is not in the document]
0xSachinK: The two other places where we are passing in `true` are module hooks. We always want to track settled funding in module hooks. Also, I can't pass in `_trackSettledFunding` as a parameter because then the functions wouldn't adhere to the `IModuleIssuanceHook` interface.

contracts/protocol/modules/PerpV2BasisTradingModule.sol
```diff
@@ -0,0 +1,511 @@
...
+     * @param _quoteBoundQuantityUnits      Max/min of vQuote asset to pay/receive when buying or selling
+     */
+    function tradeAndTrackFunding(
+        ISetToken _setToken,
+        address _baseToken,
+        int256 _baseQuantityUnits,
+        uint256 _quoteBoundQuantityUnits
+    )
+        external
+        // `trade` function has a redundancy check
```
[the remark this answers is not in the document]
0xSachinK: Oh, `redundancy` was a typo. Edited it to:
<javascript>
// Calling a `nonReentrant` function from another `nonReentrant function is not supported.
// Hence, can't use `nonReentrant` here because `PerpV2LeverageModule#trade` function has a reentrancy check.
</javascript>

contracts/protocol/modules/PerpV2BasisTradingModule.sol
0xSachinK: If we do accrue fees here, then we should first withdraw all of it and absorb it as default position and then set `settledfunding[_setToken]` to zero (similar to how we do in withdrawAndAccrueFees). But I'm not sure if that is desirable, and is what the manager excepts to happen.

contracts/protocol/modules/PerpV2BasisTradingModule.sol
```diff
@@ -0,0 +1,527 @@
...
+    function initialize(
+        ISetToken _setToken,
+        FeeState memory _settings
+    )
+        external
+    {
+        _validateFeeState(_settings);
+
+        // Initialize by calling PerpV2LeverageModule#initialize. 
+        // Verrifies caller is manager. Verifies Set is valid, allowed and in pending state.
```
richardliang, line 148: `Verifies`

contracts/protocol/modules/PerpV2BasisTradingModule.sol
```diff
@@ -0,0 +1,527 @@
...
+     * @param _quoteBoundQuantityUnits      Max/min of vQuote asset to pay/receive when buying or selling
+     */
+    function tradeAndTrackFunding(
+        ISetToken _setToken,
+        address _baseToken,
+        int256 _baseQuantityUnits,
+        uint256 _quoteBoundQuantityUnits
+    )
+        external
+        // Calling a `nonReentrant` function from another `nonReentrant` function is not supported.
```
richardliang, line 175: We can move this comment to the above javadocs

richardliang (changes requested):
Pending Brian's comments

Let's make a decision on whether `settledFunding` can be negative

contracts/protocol/modules/PerpV2BasisTradingModule.sol
bweick: In the past for things like this we always accrued fees. Though a) perhaps that's too opinionated of a stance, and b) in this case it actually updates the position state would could be a surprise so I'm ok forgoing it in this case

contracts/protocol/modules/PerpV2BasisTradingModule.sol
```diff
@@ -0,0 +1,511 @@
...
+     */
+    function _getUpdateSettledFunding(ISetToken _setToken) internal returns (uint256) {
+        // todo: Following call fetches pending funding payment across all markets. Should we change it?
+        
+        // NOTE: pendingFundingPayments are represented as in the Perp system as "funding owed"
+        // e.g a positive number is a debt which gets subtracted from owedRealizedPnl on settlement.
+        // We are flipping its sign here to reflect its settlement value.
+        int256 pendingFundingToBeSettled =  perpExchange.getAllPendingFundingPayment(address(_setToken)).neg();
+        
+        if (pendingFundingToBeSettled > 0) {
```
bweick, line 463: Should make this `>= 0` right now if pending funding is 0 then it will return 0

contracts/protocol/modules/PerpV2BasisTradingModule.sol
```diff
@@ -0,0 +1,527 @@
...
+    )
+        external
+        nonReentrant
+        onlyManagerAndValidSet(_setToken)
+    {
+        _updateSettledFunding(_setToken, _trackSettledFunding);
+
+        uint256 settledFundingInCollateralDecimals = settledFunding[_setToken].fromPreciseUnitToDecimals(collateralDecimals);
+
+        if (_notionalFunding > settledFundingInCollateralDecimals) {
```
bweick, line 216: Can we DRY this function up a bit? Put the logic for determining the `_notionalFunding` in its own function. Potentially the `collateralBalanceBeforeWithdraw` in there as well

contracts/protocol/modules/PerpV2BasisTradingModule.sol
```diff
@@ -0,0 +1,527 @@
...
+        }
+
+        uint256 collateralBalanceBeforeWithdraw = collateralToken.balanceOf(address(_setToken));
+
+        _withdraw(_setToken, _notionalFunding);
+
+        (uint256 managerFee, uint256 protocolFee) = _handleFees(_setToken, _notionalFunding);
+        
+        // Update default position unit
+        _setToken.calculateAndEditDefaultPosition(
```
bweick, line 227: Maybe group the state update logic together as well

contracts/protocol/modules/PerpV2BasisTradingModule.sol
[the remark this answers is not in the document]
0xSachinK: Good question. Realized it makes sense to auto-update it. 
Because for example, a manager can decide not to update settled funding when the pending funding is negative, which isn't a desirable outcome for investors.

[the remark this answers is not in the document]
0xSachinK: Fixed in [febb656](https://github.com/SetProtocol/set-protocol-v2/pull/198/commits/febb6562df08ceb0bda0a24bc210b5a13f0e85f0)

contracts/protocol/modules/PerpV2BasisTradingModule.sol
0xSachinK: Added in [2bafca2](https://github.com/SetProtocol/set-protocol-v2/pull/198/commits/2bafca23b9ca00e1d3899a61d00ffab68692606a)

richardliang (approved):

PR #202, opened 2022-03-11
Rebuild and publish 0.1.15

cgewecke:
Somehow the `initialize` method in typechain factory for IssuanceModule was corrupted in 0.1.14. It only contains types for two parameters (and should have 6) 

```js
{
    inputs: [
      {
        internalType: "contract ISetToken",
        name: "_setToken",
        type: "address",
      },
      {
        internalType: "contract IManagerIssuanceHook",
        name: "_preIssueHook",
        type: "address",
      },
    ],
    name: "initialize",
    outputs: [],
    stateMutability: "nonpayable",
    type: "function",
  },
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@setprotocol/set-protocol-v2",
-  "version": "0.1.14",
+  "version": "0.1.15",
   "description": "",
   "main": "dist",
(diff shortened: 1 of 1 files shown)


cgewecke:
Admin merging this (only contains a package version bump)

PR #225, opened 2022-03-18
feat(deprecation): Consolidate or remove deprecated contracts [SIM-124]

cgewecke:
(Related: [set-v2-strategies PR 22][3])

+ Create `deprecated` folder for contracts we longer need or use.
+ Deprecate NavIssuanceModule (in favor of CustomOracleNAVIssuanceModule)
+ Deprecate PerpV2LeverageModule (in favor of PerpV2LeverageModuleV2)
+ Deprecate AaveGovernanceAdapter (in favor of AaveGovernanceV2Adapter)
+ Deprecate UniswapYieldHook 
+ Deprecate all AaveV1 contracts, external ABIs and associated fixtures
+ Deprecate all single use token migration contracts

[3]: https://github.com/SetProtocol/set-v2-strategies/pull/22


bweick:
I don't think we need the `IssuanceModule` update since that was already done in a #201 by Pranav.

bweick:
In favor of deprecating all AaveV1 and token migration contracts. Maybe we can link to an example of a token migration contract in the README?

bweick (approved):

PR #229, opened 2022-03-22
chore(tests): Add BasisModule <> SlippageIssuance Integration Tests

0xSachinK:
Review suggestions
- [x] Remove duplicated tests
    - Removed duplicated `issuance` tests in 7fa3e3e
    - [Added comments](https://github.com/SetProtocol/set-protocol-v2/blob/7fa3e3e1fb3b11f8b7a7d2b87daecd87ea861d37/test/integration/perpV2BasisTradingSlippageIssuance.spec.ts#L206) to explain which parts are tested and which skipped along with reasons.
- [x] Tests for redemption when funding is non-zero (Added [here](https://github.com/SetProtocol/set-protocol-v2/blob/7fa3e3e1fb3b11f8b7a7d2b87daecd87ea861d37/test/integration/perpV2BasisTradingSlippageIssuance.spec.ts#L356) and [here](https://github.com/SetProtocol/set-protocol-v2/blob/7fa3e3e1fb3b11f8b7a7d2b87daecd87ea861d37/test/integration/perpV2BasisTradingSlippageIssuance.spec.ts#L652))
- [x] Use `getRedemptionAdjustment` in a redemption flow to test the UI flow (Added in 9b88be4)

Additions to unit tests
- [x] Setting the performance fee after withdrawing with different amounts (897a00b)
- [x] Deposit entire withdrawn amount (b46101e)


hardhat.config.ts
```diff
@@ -35,6 +35,8 @@ const config: HardhatUserConfig = {
       allowUnlimitedContractSize: false,
       forking: (process.env.FORK) ? forkingConfig : undefined,
       accounts: getHardhatPrivateKeys(),
+      gas: 12000000,
```
0xSachinK, line 38: @cgewecke Do you see any potentional bugs/errors that would slip through, due to this change? For example, we couldn't discover that the very first implementation of `BasisTradingModule` was exceeding the bytecode size limit when we had set `allowUnlitmitedContractSize` to true.

cgewecke:
(Closing and re-opening to see if that fixes semantic release bot failure)

hardhat.config.ts
cgewecke: No I don't think so. This is well under the current block sizes on the three networks we target:
+ Ethereum: 29,970,676
+ Optimism: 15,000,000
+ Polygon: 30,000,000

cgewecke (commented):
I wonder if we should use this to look at scenarios that are more specific to the Basis module. Its moduleIssuance and component hooks are identical with PerpV2LeverageModuleV2 which already runs all these tests.

Might be interesting to remove cases duplicated by the other test file and look at:
+ redemption where funding is non-zero 
+ what happens when calling methods like `withdrawFundingAndAccrueFees` 
+ using `getRedemptionAdjustments` in a redemption flow.
+ reproducing the revert Iosiro reported when depositing after a withdrawal 
+ setting the performance fee after withdrawing with different amounts. 

I know some of these are annoying to set very specific expectations for but it seems useful to exercise different scenarios and make sure there are no surprises when methods with new logic are called in combination. 

@0xSachinK What do you think?

0xSachinK:
@cgewecke Added in 9b88ber

0xSachinK:
Rebased to master.

test/integration/perpV2BasisTradingSlippageIssuance.spec.ts
```diff
@@ -231,554 +203,19 @@ describe("PerpV2BasisTradingSlippageIssuance", () => {
...
-      });
-    });
-  });
-
+  // PerpV2BasisTradingModule#moduleIssueHook implementation calls PerpV2LeverageModuleV2#moduleIssueHook to handle issuance
+  // after updating tracked settled funding. The functionality to track settled funding when moduleIssueHook is called is
+  // tested in the PerpV2BasisTradingModule unit tests (test/protocol/modules/perpV2BasisTradingModule.spec.ts). And the
+  // functionality to set external position unit before issuance, has been tested in the PerpV2LeverageModuleV2 <> SlippageIssuanceModule
+  // integration tests (test/integration/perpV2LeverageV2SlippageIssuance.spec.ts). Hence we will not restest them here.
+  describe("#issuance", () => {});
```
cgewecke, line 211: 👍

cgewecke (commented):
This is looking really good. I think the only remaining question is about the revert Iosiro triggered when depositing after a withdrawal. They suggested we airdrop into the Set:

> was able to recreate the issue without any withdrawing and just deposit with airdropped tokens: https://gist.github.com/bernard-wagner/a1e8783404256a898e4df66d2b0a83ed

> Seems to be that calculateAndEditDefaultPosition does not handle airdropped tokens correctly when the balance has been reduced to less than the airdropped amount. 

> This happens more often with BasisTradingModule as the default position is rounded down in calculateDefaultEditPositionUnit when withdrawing. Thus the difference between notionalAmount and totalSupply*positionUnit is regarded as airdropped tokens when depositing. 

WDYT about this and should we consider using Airdrop module or something to manage it?

PR #242, opened 2022-04-08
fix(tests): Add Hardhat deployment subgraph test scripts 

comc:
Local Hardhat node state deployment scripts for subgraph development and testing have been migrated from the set-protocol-v2-subgraph repo to a new `subgraph/test/` directory in this repo, as these scripts are dependent on a Hardhat chain deployment from this repo.


cgewecke (approved):
It might be nice to add a comment in these scripts about their use context. 

It it correct that this is intended to be used at [set-protocol-v2-subgraph][1] with a command like?
```
yarn deploy-hardhat $PWD/node_modules/@setprotocol/set-protocol-v2/subgraph/test/deploy-state-multi-token.ts
```




[1]: https://github.com/SetProtocol/set-protocol-v2-subgraph

PR #244, opened 2022-04-13
fix(PerpV2BasisTradingModule): Expose getter to fetch updated settled funding

0xSachinK:
This getter would be consumed by the basis trading strategy extension contract to determine the amount that is available to be reinvested.


cgewecke (approved):
LGTM! 🏅 

(It looks like coveralls is having some delays reporting results - verified that the tests are hitting relevant lines locally as below)

Screen Shot 2022-04-13 at 5 55 43 AM

PR #251, opened 2022-05-02
Notional Trade Module

ckoopmann:
# Code Review Processes
## New Feature Review
* Design doc is created and posted here: [https://hackmd.io/@r1OYOYxYSP-6k15CBaosBQ/rJ2jgYU4q]

README Checks

Code Checks
- [x] Add javadocs. 

Broader Considerations


contracts/protocol/modules/v1/NotionalTradeModule.sol
```diff
@@ -0,0 +1,598 @@
...
+        IWrappedfCashComplete _fCashPosition,
+        uint256 _fCashAmount,
+        bool _toUnderlying
+    )
+    internal
+    {
+        // TODO: Review if this value is correct / what is max implied rate ? 
+        uint32 maxImpliedRate = type(uint32).max;
+        // TODO: This is copied from the notional sdk proxy. Review if this is really more efficient that just invoking "burn" through the setToken (as in the mint case
+        IERC777(address(_fCashPosition)).operatorBurn(
```
jeffywu, line 500: I think you should be able to invoke from the contract holding the tokens rather than using an operator. That feels safer to me.

contracts/protocol/modules/v1/NotionalTradeModule.sol
```diff
@@ -0,0 +1,598 @@
...
+        address sendToken,
+        uint256 preTradeSendTokenBalance,
+        address receiveToken,
+        uint256 preTradeReceiveTokenBalance
+    ) internal returns (uint256, uint256) {
+
+        uint256 setTotalSupply = setToken.totalSupply();
+
+        // This reverts if i try to trade a token that is not a registered component. (I.e. I just sent a lot of sendToken to the contract manually and then try to trade it.
+        // TODO: Review if this needs to be addressed
```
jeffywu, line 579: I didn't see this in the contract but might be missing it, when you roll to a new maturity you need to make sure a wrapper is deployed via the factory first.

package.json
```diff
@@ -105,11 +105,13 @@
     "web3": "^1.2.9"
   },
   "dependencies": {
+    "@uniswap/v3-core": "1.0.1",
     "@uniswap/v3-sdk": "^3.5.1",
     "ethers": "^5.5.2",
     "fs-extra": "^5.0.0",
     "jsbi": "^3.2.5",
     "module-alias": "^2.2.2",
+    "notional-solidity-sdk": "https://github.com/ckoopmann/notional-solidity-sdk.git#77aa2f8fd8a72b8f5865e561f2d39d4606c3740c",
```
jeffywu, line 114: I've pulled just the necessary code for this project into its own repo: https://github.com/notional-finance/wrapped-fcash

contracts/protocol/modules/v1/NotionalTradeModule.sol
ckoopmann: I agree, adjusted accordingly.

package.json
ckoopmann: Thanks for the info, switched the dependency out accordingly. 
Note that, I will likely remove the depencency as well as all of the deploy / upgrade logic once we have a mainnet deployment of the fcash wrapper

contracts/protocol/modules/v1/NotionalTradeModule.sol
```diff
@@ -0,0 +1,598 @@
...
+        ISetToken _setToken,
+        IWrappedfCashComplete _fCashPosition,
+        uint256 _maxAssetAmount,
+        uint256 _fCashAmount,
+        bool _fromUnderlying
+    )
+    internal
+    {
+        // TODO: Review if this value is correct / what is min implied rate ? 
+        uint32 minImpliedRate = 0;
```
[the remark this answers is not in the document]
ckoopmann: We are already letting the user, specify a maximum amount of underlying / asset token to spend. Doesn't that already provide the respective slippage / MEV protection ?

contracts/protocol/modules/v1/NotionalTradeModule.sol
```diff
@@ -0,0 +1,598 @@
...
+    function _burn(
+        ISetToken _setToken,
+        IWrappedfCashComplete _fCashPosition,
+        uint256 _fCashAmount,
+        bool _toUnderlying
+    )
+    internal
+    {
+        // TODO: Review if this value is correct / what is max implied rate ? 
+        uint32 maxImpliedRate = type(uint32).max;
```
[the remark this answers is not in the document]
ckoopmann: Question as before, when already supplying a minimum of asset / underlying token to receive do we need this parameter for slippage / MEV protection ?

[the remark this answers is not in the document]
ckoopmann: As I read here you can calculate one from the other, that's why I'm confuse why we would have to specify both
https://docs.notional.finance/developer-documentation/how-to/lend-and-borrow-fcash

contracts/protocol/modules/v1/NotionalTradeModule.sol
ckoopmann: As per the spec ( https://hackmd.io/@r1OYOYxYSP-6k15CBaosBQ/rJ2jgYU4q ) trades in and out of existing wrappedFCash tokens. Also at the time of maturity we only redeem the matured positions to the cToken which is to be reinvested later manually. So according to my understanding we wouldn't need to deploy the wrappers from this contract directly.

I will verify with set though wether that understanding is correct.

contracts/protocol/modules/v1/NotionalTradeModule.sol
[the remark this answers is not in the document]
jeffywu: Yes it does, the implied rate is just a way to do it without having to rely on the block time. But your way is just as good.

contracts/protocol/modules/v1/NotionalTradeModule.sol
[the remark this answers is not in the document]
jeffywu: You do not, similar to answer above. You can also use 0 for this to not take effect, but your solution will work fine as well.

contracts/protocol/modules/v1/NotionalTradeModule.sol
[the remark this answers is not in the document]
ckoopmann: Thanks, will keep it as is then.

cgewecke (commented):
@ckoopmann This looks really nice! Integration tests look solid. 

Have left some (mostly nit) comments and a couple smaller questions.  

(Would it be possible to regenerate the abi JSON files added in this PR so they don't add 200k lines? They can't be inspected on Github and had trouble opening them locally as well.)

ckoopmann:
> @ckoopmann This looks really nice! Integration tests look solid.
> 
> Have left some (mostly nit) comments and a couple smaller questions.
> 
> (Would it be possible to regenerate the abi JSON files added in this PR so they don't add 200k lines? They can't be inspected on Github and had trouble opening them locally as well.)

This is mostly the bytecode for the notional proxy update shared with me by @jeffywu. Maybe he can direct me to the smart contract source code from which that bytecode was generated so we can remove all those bytecode from the repo.

contracts/protocol/modules/v1/NotionalTradeModule.sol
```diff
@@ -0,0 +1,641 @@
...
+    function _getPaymentToken(
+        IWrappedfCashComplete _fCashPosition,
+        bool _getUnderlying
+    )
+    internal
+    view
+    returns(IERC20 paymentToken)
+    {
+        (IERC20 underlyingToken, IERC20 assetToken) = _getUnderlyingAndAssetTokens(_fCashPosition);
+         paymentToken = _getUnderlying ? underlyingToken : assetToken;
```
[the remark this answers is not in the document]
ckoopmann: Ah, good point. Will add a test case where we toggle the redeemToUnderlying setting accordingly.

contracts/interfaces/external/INotionalProxy.sol
```diff
@@ -0,0 +1,4 @@
+pragma solidity ^0.8.11;
```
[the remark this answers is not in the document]
ckoopmann: Thanks for the hint, done.

contracts/protocol/modules/v1/NotionalTradeModule.sol
```diff
@@ -0,0 +1,641 @@
...
+    /**
+     * @dev MANAGER ONLY: Removes this module from the SetToken, via call by the SetToken. Redeems any matured positions
+     */
+    function removeModule() external override onlyValidAndInitializedSet(ISetToken(msg.sender)) {
+        ISetToken setToken = ISetToken(msg.sender);
+
+        // Redeem matured positions prior to any removal action
+        _redeemMaturedPositions(setToken);
+
+        delete fCashPositions[setToken];
```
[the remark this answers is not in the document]
ckoopmann: I expect it to do so, will double check.

contracts/mocks/WrappedfCashMock.sol
```diff
@@ -0,0 +1,145 @@
+/*
```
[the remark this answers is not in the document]
ckoopmann: Great point thanks, adjusted accordingly.

contracts/protocol/modules/v1/NotionalTradeModule.sol
```diff
@@ -0,0 +1,641 @@
...
+            if(fCashPosition.hasMatured()) {
+                IERC20 receiveToken = _getPaymentToken(fCashPosition, toUnderlying);
+                uint256 fCashBalance = fCashPosition.balanceOf(address(_setToken));
+                _redeemFCashPosition(_setToken, fCashPosition, receiveToken, fCashBalance, 0);
+                fCashPositions[_setToken].remove(address(fCashPosition));
+            }
+
+        }
+
+    }
```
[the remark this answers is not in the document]
ckoopmann: Fixed

contracts/protocol/modules/v1/NotionalTradeModule.sol
```diff
@@ -0,0 +1,641 @@
...
+import { IController } from "../../../interfaces/IController.sol";
+import { IDebtIssuanceModule } from "../../../interfaces/IDebtIssuanceModule.sol";
+import { IModuleIssuanceHook } from "../../../interfaces/IModuleIssuanceHook.sol";
+import { IWrappedfCash, IWrappedfCashComplete } from "../../../interfaces/IWrappedFCash.sol";
+import { ISetToken } from "../../../interfaces/ISetToken.sol";
+import { ModuleBase } from "../../lib/ModuleBase.sol";
+import "@openzeppelin/contracts/utils/EnumerableSet.sol";
+
+
+contract NotionalTradeModule is ModuleBase, ReentrancyGuard, Ownable, IModuleIssuanceHook {
```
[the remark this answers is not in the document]
ckoopmann: Added a docstring similar in format / length to the other modules, lmk in case this is not "long" enough. (I thought the formulation in the spec was a bit to verbose for a code comment)

contracts/protocol/modules/v1/NotionalTradeModule.sol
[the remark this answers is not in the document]
ckoopmann: Actually I was even missing a method for the manager to change this parameter, which I added now

contracts/protocol/modules/v1/NotionalTradeModule.sol
```diff
@@ -0,0 +1,641 @@
...
+     * @param _fCashPositions             Addresses of fCash wrappers to register for trading / maturity redemption
+     */
+    function addFCashPositions(
+        ISetToken _setToken,
+        address[] calldata _fCashPositions
+    )
+    external
+    onlyManagerAndValidSet(_setToken)
+    {
+        _addFCashPositions(_setToken, _fCashPositions);
```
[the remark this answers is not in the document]
ckoopmann: I completely removed the registered fCash positions and instead used the set tokens positions directly combined with a function "isWrappedFCash" that checks if an address is an fcash wrapper deployed from the factory.

contracts/protocol/modules/v1/NotionalTradeModule.sol
ckoopmann: I rewrote parts of the contract and changed the trade methods such that when trading into a wrapped fCash position it uses the factory to deploy the wrapper if it doesn't exist yet.

contracts/protocol/modules/v1/NotionalTradeModule.sol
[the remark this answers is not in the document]
ckoopmann: This topic is not relevant anymore since I removed the `fCashPositions` mapping entirely.

contracts/protocol/modules/v1/NotionalTradeModule.sol
```diff
@@ -0,0 +1,692 @@
...
+
+    /**
+     * @dev Checks if a given address is an fCash position that was deployed from the factory
+     */
+    function _isWrappedFCash(address _fCashPosition) internal view returns(bool){
+        if(!_fCashPosition.isContract()) {
+            return false;
+        }
+        uint16 currencyId;
+        try IWrappedfCash(_fCashPosition).getCurrencyId() returns(uint16 _currencyId){
```
jeffywu, line 638: There is a method called getDecodedID that will return both currencyId and maturity in a single call

contracts/protocol/modules/v1/NotionalTradeModule.sol
```diff
@@ -0,0 +1,692 @@
...
+            // Check that the given position is an equity position
+            if(positions[i].unit > 0) {
+                address component = positions[i].component;
+                if(_isWrappedFCash(component)) {
+                    numFCashPositions++;
+                }
+            }
+        }
+
+        // TODO: Looping twice just to get the array length for initialization feels inefficient, review if there is a better way.
```
jeffywu, line 613: The loop twice to get the length thing is something I've had to do as well. I'm not sure how generic this module is going to be, but if you're really only going to have fCash + 1 token that you redeem to (i.e. cUSDC or cDAI) then it might be okay just to do array length - 1 here.

Again, not sure how much that really saves you vs. making it less future proof. Probably something to discuss with Allan but I imagine even if you did a FIXED product with 2 currencies (i.e. fDAI and fUSDC) you might still want to only have redemptions in 1 currency (either cUSDC or cDAI) for simplicity. But then again, I don't know.

A couple other options you have (maybe) is to check for address(0) as the termination point of the array.

Or you could do this:
https://ethereum.stackexchange.com/a/51897/1148

contracts/protocol/modules/v1/NotionalTradeModule.sol
```diff
@@ -0,0 +1,692 @@
...
+     */
+    function _getPaymentToken(
+        IWrappedfCashComplete _fCashPosition,
+        bool _getUnderlying
+    )
+    internal
+    view
+    returns(IERC20 paymentToken)
+    {
+        (IERC20 underlyingToken, IERC20 assetToken) = _getUnderlyingAndAssetTokens(_fCashPosition);
```
jeffywu, line 575: You could put an if statement in here and just call the view method on the wrapper that you need, I think it should save you a bit of gas.

```
if (_getUnderlying) {
  (paymentToken,) = _fCashPosition.getUnderlyingToken()
} else {
  (paymentToken,) = _fCashPosition.getAssetToken()
}
```

Also there is a getToken method on the wrapper that basically does this for you:
https://github.com/notional-finance/wrapped-fcash/blob/master/contracts/wfCashBase.sol#L139

contracts/protocol/modules/v1/NotionalTradeModule.sol
```diff
@@ -0,0 +1,692 @@
...
+        IWrappedfCashComplete _fCashPosition,
+        uint256 _fCashAmount,
+        bool _toUnderlying
+    )
+    internal
+    {
+        uint32 maxImpliedRate = type(uint32).max;
+
+        string memory functionSignature =  
+            _toUnderlying ? "redeemToUnderlying(uint256,address,uint32)": "redeemToAsset(uint256,address,uint32)";
```
jeffywu, line 535: I'm not 100% sure on this but I think using selectors will be more gas efficient here since you don't have to load the string into memory, the bytes4 will just stay on the stack. I don't think it will hurt the readability of the code since you can do something like:

```
bytes4 functionSelector = _toUnderlying ? IWrappedfCashComplete.redeemToUnderlying.selector : IWrappedfCashComplete.redeemToAsset.selector;
... abi.encodeWithSelector( ... );
```

contracts/protocol/modules/v1/NotionalTradeModule.sol
```diff
@@ -0,0 +1,692 @@
...
+    function _approveIfNecessary(
+        ISetToken _setToken,
+        IWrappedfCashComplete _fCashPosition,
+        IERC20 _sendToken,
+        uint256 _maxAssetAmount
+    )
+    internal
+    {
+        if(IERC20(_sendToken).allowance(address(_setToken), address(_fCashPosition)) < _maxAssetAmount) {
+            // TODO: Review if we want to only approve "_maxAssetAmount" or keep it at maxUint256
```
jeffywu, line 488: If using the mintViaAsset and mintViaUnderlying methods, the fCash wrapper will only ever transfer `depositAmountExternal` so you can use this as the approval limit.

https://github.com/notional-finance/wrapped-fcash/blob/master/contracts/wfCashLogic.sol#L59

contracts/protocol/modules/v1/NotionalTradeModule.sol
jeffywu: It looks like this is just used inside an external view so I guess efficiency is not a huge requirement.

contracts/protocol/modules/v1/NotionalTradeModule.sol
```diff
@@ -0,0 +1,692 @@
...
+
+    /**
+     * @dev Returns an array with fcash position addresses for given set token
+     */
+    function _getFCashPositions(ISetToken _setToken)
+    internal
+    view
+    returns(address[] memory fCashPositions)
+    {
+        ISetToken.Position[] memory positions = _setToken.getPositions();
```
jeffywu, line 599: Looking at this file I don't see where the setToken actually registers the different positions, I guess maybe that happens in a different module?

cgewecke (commented):
@ckoopmann Went back over this and things LGTM. I see jeffywu left a few additional notes you might address so will hold off on final approval until those are resolved.

RE: the abi's .. I'm mostly concerned about readability but your present solution lgtm if these will eventually be removed.

In terms of next steps... there's quite a bit of complexity here - it would be good to test this module as a staging-mainnet environment deployment before launching anything and make sure a) it works and b) nothing critical is missing in terms of being able to represent the asset's yield on the IC website. 

(I see there are a couple of TODO notes in the `#mint/redeemFCashPosition` integration tests about how to calculate values returned from the Notional exchange when minting that would be nice to resolve before this goes to production as well.)

Nice work!

contracts/protocol/modules/v1/NotionalTradeModule.sol
```diff
@@ -0,0 +1,692 @@
...
+     * @param _wrappedfCashFactory              Address of fCash wrapper factory used to check and deploy wrappers
+     */
+    constructor(
+        IController _controller,
+        IWrappedfCashFactory _wrappedfCashFactory
+    )
+        public
+        ModuleBase(_controller)
+    {
+        // TODO: Review if we want to make this address mutable. (made it immutable initially to be consistent with other modules)
```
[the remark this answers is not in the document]
ckoopmann: Thanks for the confirmation.

contracts/protocol/modules/v1/NotionalTradeModule.sol
ckoopmann: Yeah, it's only a view so shouldn't make much of a difference except in contract deployment costs.

I just wanted to make sure I wasn't missing anything obvious. I will keep it as is I think. 

Thanks a lot for the suggestions.

contracts/protocol/modules/v1/NotionalTradeModule.sol
ckoopmann: Good call, thanks. Refactored accordingly

contracts/protocol/modules/v1/NotionalTradeModule.sol
ckoopmann: Good point thanks. Totally missed that. I removed this helper method and used the `getToken` method on the wrapper instead.

contracts/protocol/modules/v1/NotionalTradeModule.sol
ckoopmann: Positions are adjusted after minting / redeeming using this function:
https://github.com/ckoopmann/set-protocol-v2/blob/ca922c42e2f9065e2bd0964507e49edf28ab185f/contracts/protocol/modules/v1/NotionalTradeModule.sol#L637

This will adjust the position amount and also add / remove fCash components when necessary.

contracts/protocol/modules/v1/NotionalTradeModule.sol
ckoopmann: Great point. I was actually thinking about this, delayed it and then never got back to it.

I replaced all string signatures with selectors which seems to save some gas (according to  my very superficial analysis) but also feels more elegant to me.

cgewecke (approved):
LGTM!

PR #252, opened 2022-05-03
fix(interfaces): Add IAirdropModule, IClaimModule, and IWrapModuleV2 [SIM-270]

pblivin0x:
To provide necessary interfaces for [WrapExtension and ClaimExtension ](https://github.com/SetProtocol/set-v2-strategies/pull/35)


cgewecke (approved):
Perfect 💯

PR #253, opened 2022-05-13
Support Curve stableswap pools

FlattestWhite:
# Code Review Processes

Curve pools can have 2, 3 or even 4 coins in a pool. However, from an exchange perspective, we'd only ever be swapping two coins at one time. Furthermore, most coins in all pools are ERC20 tokens. Hence the eth/steth pool needed some extra logic around wrapping and unwrapping for eth/weth. The approach taken with this adapter is to assume we're only going to be trading ERC20 tokens. 

It assumes both tokens that are being traded are ERC20 tokens and it requires the index on of the coin on the curve pool to be passed in at construction. The rest is just a simple encoding of the exchange function.

Newer curve pools also have an exchange function that can take a destination address as the last argument. This means we can save an extra contract call.  However, most existing curve pools don't support that new exchange method. Hence, I've decided to use the older more widesperad "exchange" method, but it requires doing an extra transfer of tokens back to the user.

## New Feature Review

README Checks

Code Checks
- [] Add javadocs. 

Broader Considerations


contracts/mocks/external/CurveStableswapMock.sol
```diff
@@ -56,18 +65,19 @@ contract CurveStEthStableswapMock is ReentrancyGuard {
...
-        if (_i == 0 && _j == 1) {
-        // The caller has sent eth receive stETH
-        require(_dx == msg.value);
-        IERC20(tokens[1]).safeTransfer(msg.sender, _dx);
-        } else if (_j == 0 && _i == 1) {
-        // The caller has sent stETH to receive ETH
-        IERC20(tokens[1]).safeTransferFrom(msg.sender, address(this), _dx);
-        Address.sendValue(msg.sender, _dx);
+
+        if (tokens[uint256(_i)] == ETH_TOKEN_ADDRESS) {
```
FlattestWhite, line 69: Support eth transfers for mocking

cgewecke (commented):
@FlattestWhite This looks really good.

Will this also integrate with the AaveLeverageModule? If so is it possible to copy and adapt the existing forked integration tests for the CurveStEth adapter to validate the behavior of this contract?

Have tried this locally as part of the review and a couple of its tests are failing and giving results that I don't understand (probably my mistake). The lever tests for:
+ "should update the collateral position on the SetToken correctly"
+ "should transfer the correct components from the exchange" 
have broken expectations. 

That doesn't mean anything's incorrect here ... would just like the test suite to have a good description of how the adapter is expected to behave in the wild. 

You'll need to: 
+ update the hardhat config's forking block number to something more recent since the WETH stableswap pool you've added to the dependencies is newish
+ make an adjustment to the existing Uniswap adapter forked integration tests to reflect current price conditions as Christian has done [in his notional PR here][1])

[1]:  https://github.com/SetProtocol/set-protocol-v2/pull/251/files#diff-68002b9f810ff8f9e2c245c8eeeeaa35d217d65c250d9862be31015ea73506dd

FlattestWhite:
> LGTM! Thanks for the additional tests and the clarifying comments - very helpful.

Thanks - will your team merge and write the deploy script? Or will you need me to do so?

cgewecke:
@FlattestWhite Just merged - we will deploy the adapter. I've asked some questions about how to parameterize the deployment in a slack channel Set shares with IC which Ed added you to yesterday. 

Will continue discussion there :)

PR #255, opened 2022-05-25
Rename WrappedfCashFactory to WrappedfCashFactoryExperimental

cgewecke:
As part of #251, several contracts that will be published/deployed by Notional in future have been temporarily brought into the repo to enable NotionalTradeModule development and staging_mainnet testing

[WrappedfCashFactory][1] is one of these. Unfortunately the way it's imported results in duplicate hardhat artifacts being generated and two problems in the downstream deployments repo:
+ hardhat-etherscan errors with a complaint that it cannot disambiguate between the contract bytecodes for verification
+ hardhat-deploy errors with a complaint that it cannot locate the contract for deployment 

This PR renames the root-level contract and makes its bytecode slightly different (in an irrelevant way) to get around these.

[1]: https://github.com/SetProtocol/set-protocol-v2/blob/master/contracts/protocol/integration/wrap/notional/WrappedfCashFactory.sol


ckoopmann:
Note that https://github.com/SetProtocol/set-protocol-v2/pull/256 will remove these temporarily added contracts and should thereby fix these issues.

PR #260, opened 2022-07-21
ArrakisUniswapV3AmmAdapter

zishansami102:
The changes implement an adapter for the AmmModule which supports Arrakis Vaults holding UniswapV3 positions.

The goal is to be able to create and hold Uniswap's V3 Liquidity Provision tokens (LP tokens) inside of a Set Token.
There is a challenge caused by UniV3's LP tokens being ERC721s (NFTs). And since `AmmModule.sol` assumes amm pool tokens to follow IERC20, UniswapV3 Pools aren't directly integrable. 

So are using Arrakis ([Arrakis V1 Doc](https://docs.arrakis.fi/developer-docs#introduction-arrakis-v1)) that creates and holds Uniswap V3 Liquidity Position and issues fungible ERC20 tokens to anyone who adds more liquidity.

There are three main contracts being used in the implementation:
1. Arrakis Facotry - [ArrakisFactoryV1.sol](https://github.com/ArrakisFinance/vault-v1-core/blob/main/contracts/ArrakisFactoryV1.sol)
2. Arrakis Vault - [ArrakisVaultV1.sol](https://github.com/ArrakisFinance/vault-v1-core/blob/main/contracts/ArrakisVaultV1.sol)
3. Arrakis Router - GUniRouter.sol - [Mainnet Deployment](https://etherscan.io/address/0x9de3f4847186677e51cc9e39e07d30bd58927f6a#code)

Other Relevant Contracts - [Link](https://docs.arrakis.fi/pool-addresses)

Using Arrakis Factory, anyone can create new Arrakis Vaults (analogous to Token Pools) which hold UniswapV3 liquidity position for a pair of ERC20 tokens for a given price range bound (assuming there already exists a pool for the same pair of tokens in UniswapV3).

Once Vault is created, it works similar to UniswapV2 liquidity pools and can be passed to `AmmModule` contract as ammPool, which then will call `ArrakisUniswapV3Adapter` returning calldata with Arrakis Router, which handles adding and removing liquidity and minting/burning ERC20 Tokens to the caller (in our case it is SetToken).

The `router` in the `ArrakisUniswapV3Adapter.sol` is the address to the Arrakis Router contract which is passed in the constructor along with UniswapV3 Factory address. The overall approach is now very similar to how `UniswapV2AmmAdapter` is implemented with the following correlations:
 - Arrakis Facotry - UniswapV2 Factory
 - Arrakis Vault - UniswapV2 Liquidity Pools
 - Arrakis Router - UniswapV2 Router


contracts/protocol/integration/amm/ArrakisUniswapV3AmmAdapter.sol
```diff
@@ -0,0 +1,281 @@
...
+        return true;
+    }
+
+    /**
+     * Sorts the amount in order of tokens stored in Arrakis/UniswapV3 Pool
+     *
+     * @param  _token0        Address of token0
+     * @param  _token0        Address of token1
+     * @param  _amount0       Amount of token0
+     * @param  _token1        Amount of token1
```
FlattestWhite, line 276:
```suggestion
     * @param  _amount1        Amount of token1
```

FlattestWhite (approved):
lgtm

contracts/protocol/integration/amm/ArrakisUniswapV3AmmAdapter.sol
```diff
@@ -0,0 +1,281 @@
...
+contract ArrakisUniswapV3AmmAdapter is IAmmAdapter {
+    using SafeMath for uint256;
+
+    /* ============ State Variables ============ */
+
+    // Address of Arrakis Router contract
+    address public immutable router;
+
+    // UniswapV3 factory contract
+    IUniswapV3Factory public immutable uniV3Factory;
```
FlattestWhite, line 44: To remove unused variable

contracts/protocol/integration/amm/ArrakisUniswapV3AmmAdapter.sol
```diff
@@ -0,0 +1,281 @@
...
+    function isValidPool(address _pool, address[] memory _components)
+        external
+        view
+        override
+        returns (bool)
+    {
+        // Attempt to get the tokens of the provided pool
+        address token0;
+        address token1;
+        try IArrakisVaultV1(_pool).token0() returns (IERC20 _token0) {
```
ckoopmann, line 235: Try catch statements like these don't catch the case when `_pool` is not a smart contract, in which case the whole tx will still fail. If you want to avoid that you might want to add a codesize check for the `_pool` address.

See this so question that I posted regarding this topic:
https://ethereum.stackexchange.com/questions/128845/why-does-soliditys-try-catch-not-catch-call-to-non-contract-address-revertion

contracts/protocol/integration/amm/ArrakisUniswapV3AmmAdapter.sol
```diff
@@ -0,0 +1,281 @@
...
+
+    // Address of Arrakis Router contract
+    address public immutable router;
+
+    // UniswapV3 factory contract
+    IUniswapV3Factory public immutable uniV3Factory;
+
+    // Internal function string for adding liquidity
+    string internal constant ADD_LIQUIDITY =
+        "addLiquidity(address,uint256,uint256,uint256,uint256,address)";
```
ckoopmann, line 48: Just checking, we don't include `addLiquidityEth` interface in this adapter since we only allow providing liquidity with ERC20 and not native ETH correct ?

contracts/protocol/integration/amm/ArrakisUniswapV3AmmAdapter.sol
```diff
@@ -0,0 +1,281 @@
...
+        address /*_component*/,
+        uint256 /*_maxTokenIn*/,
+        uint256 /*_minLiquidity*/
+    )
+        external
+        view
+        override
+        returns (address /*target*/, uint256 /*value*/, bytes memory /*data*/)
+    {
+        revert("Arrakis single asset addition is not supported");
```
ckoopmann, line 138: Couldn't this be implemented using the `rebalanceAndAddLiquidity` method on the router ? Any specific reason why we didn't implement this ?

ckoopmann (commented):
LGTM. Mostly questions for my own understanding

